PR #1701, opened 2019-12-12
Added FastTryLock class

sergey-chernikov:


sergey-chernikov:
Closing, as no demand in it. Branch will be floating around for a while (just in case).

PR #1873, opened 2020-04-16
Add signer visibility messages

pavel-kokolemin:
http://185.213.153.35:8081/browse/BST-2614


Blocksettle_proto/Communication/bs_signer.proto
```diff
@@ -44,6 +44,7 @@ enum PacketType
    SyncSettingsRequestType = 25;
 
    UpdateStatusType   = 26;
+   WindowStatusType   = 3;  // Sends headless::WindowStatus
```
sergey-chernikov, line 47: Since it's anyway put to the middle of other types, I suggest to order it properly: after 2 and before 4

pavel-kokolemin: Sure, fixed

PR #1887, opened 2020-04-20
Add ledger support

romaniy173:
Issue: add changes need to implement ledger logic in solution


WalletsLib/CoreHDGroup.h
```diff
@@ -65,6 +65,7 @@ namespace bs {
 
             virtual void shutdown(void);
             virtual std::set<AddressEntryType> getAddressTypeSet(void) const;
+            std::set<AddressEntryType> getAddressTypeHWSet(void) const;
```
sergey-chernikov, line 68: I think this should be the other way around: you descend a group for HW wallets and override getAddressTypeSet there. Don't duplicate each method for HW wallet.

WalletsLib/HDPath.h
```diff
@@ -55,10 +58,15 @@ namespace bs {
...
 
+         std::vector<Elem>::const_iterator begin() const;
+         std::vector<Elem>::const_iterator end() const;
+
          static Path fromString(const std::string &);
          static Elem keyToElem(const std::string &key);
          static std::string elemToKey(Elem);
 
+
+
```
sergey-chernikov, line 69: These 2 empty lines are not needed here, IMO.

WalletsLib/CoreHDGroup.h
romaniy173: Fixed, thank you

WalletsLib/HDPath.h
romaniy173: Fixed, thank you

WalletsLib/CoreHDWallet.cpp
```diff
@@ -313,7 +313,11 @@ void bs::core::hd::Wallet::createHwStructure(const bs::core::wallet::HwWalletInf
       { AddressEntryType_P2PKH, walletInfo.xpubLegacy_ }
    };
 
-   for (const auto &aet : groupXBT->getAddressTypeHWSet()) {
+   // Adding legacy address support explicitly
+   auto addressTypeSet = groupXBT->getAddressTypeSet();
+   addressTypeSet.insert(AddressEntryType_P2PKH);
```
sergey-chernikov, line 318: It's done wrong again. You need an HWGroup class instance where you override getAddressTypeSet method to contain the desirable address types for HWGroup only.

romaniy173: Thank you, now I got the idea. Done

WalletsLib/CoreHDWallet.cpp
```diff
@@ -222,27 +222,34 @@ std::shared_ptr<hd::Group> hd::Wallet::createGroup(bs::hd::CoinType ct)
...
+         break;
+
+      default:
+         result = std::make_shared<Group>(
+            walletPtr_, ct, netType_, extOnlyFlag_, logger_);
+         break;
+      }
+   }
+   else {
+      result = std::make_shared<HWGroup>(
```
sergey-chernikov, line 249: I guess you should forbid the creation of groups for unsupported coin types in HW wallets (e.g. Auth and CC).

sergey-chernikov: Ideally, you should subclass bs::core::hd::Wallet for HW wallet to eliminate all these `if`s and duplication of methods.

romaniy173: Agree with both. I have added exception like a fix for a first comment. The second one is more bigger, we also do not have correct structure for cases like readonly wallet etc, so everything is managing on just flag for now. So I would say I would skip second one and probably will create a ticket to further refactoring of this structure, as far as it not pretty straight forward?

PR #1894, opened 2020-04-22
Allow finding leaf by purpose and change tx input interface

romaniy173:
Issue:
1. Add function to get leaf by purpose
2. Disallow to create SelectedTransactionInputs with group, to make possible to filter out leafs from it

### DIFF
--- a/BlocksettleNetworkingLib/TransactionData.cpp
+++ b/BlocksettleNetworkingLib/TransactionData.cpp
@@ -55,6 +55,7 @@ bool TransactionData::setWallet(const std::shared_ptr<bs::sync::Wallet> &wallet
    }
    if (wallet != wallet_) {
       wallet_ = wallet;
+      group_ = nullptr;
 
       selectedInputs_ = std::make_shared<SelectedTransactionInputs>(wallet_
          , isSegWitInputsOnly_, confirmedInputs_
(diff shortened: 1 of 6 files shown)


BlocksettleNetworkingLib/TransactionData.cpp
```diff
@@ -86,25 +87,32 @@ bool TransactionData::setWallet(const std::shared_ptr<bs::sync::Wallet> &wallet
 }
 
 bool TransactionData::setGroup(const std::shared_ptr<bs::sync::hd::Group> &group
-   , uint32_t topBlock, bool resetInputs, const std::function<void()> &cbInputsReset)
+   , uint32_t topBlock, bool excludeLegacy, bool resetInputs, const std::function<void()> &cbInputsReset)
 {
    if (!group) {
       return false;
    }
+   std::vector<std::shared_ptr<bs::sync::Wallet>> wallets;
```
sergey-chernikov, line 95: This should go under `group != group_` condition, I guess

BlocksettleNetworkingLib/Wallets/SyncHDGroup.cpp
```diff
@@ -37,6 +37,17 @@ std::shared_ptr<hd::Leaf> hd::Group::getLeaf(const bs::hd::Path &path) const
    return itLeaf->second;
 }
 
+std::shared_ptr<bs::sync::hd::Leaf> bs::sync::hd::Group::getLeaf(bs::hd::Purpose purpose) const
+{
+   for (auto& leaf : getLeaves()) {
```
sergey-chernikov, line 42: const qualifier is advised here

BlocksettleNetworkingLib/TransactionData.cpp
romaniy173: No, I need also it in other condition when else if (resetInputs) {

BlocksettleNetworkingLib/Wallets/SyncHDGroup.cpp
romaniy173: fixed, thank you

PR #1896, opened 2020-04-22
Fix mixing segwit support for ledger

romaniy173:
Issue: add support to signer to allow mixing native and nested segwit for ledger tx

### DIFF
--- a/BlocksettleNetworkingLib/HeadlessContainer.cpp
+++ b/BlocksettleNetworkingLib/HeadlessContainer.cpp
@@ -458,8 +458,8 @@ bs::signer::RequestId HeadlessContainer::signMultiTXRequest(const bs::core::wall
 
    headless::SignTXMultiRequest request;
    for (const auto &input : txMultiReq.inputs) {
-      request.add_walletids(input.walletId_);
-      signer.addSpender(std::make_shared<ScriptSpender>(input.utxo_));
+      request.add_walletids(input.walletId);
+      signer.addSpender(std::make_shared<ScriptSpender>(input.utxo));
    }
    for (const auto &recip : txMultiReq.recipients) {
       signer.addRecipient(recip);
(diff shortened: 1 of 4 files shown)


BlocksettleNetworkingLib/HeadlessContainer.cpp
```diff
@@ -458,8 +458,8 @@ bs::signer::RequestId HeadlessContainer::signMultiTXRequest(const bs::core::wall
 
    headless::SignTXMultiRequest request;
    for (const auto &input : txMultiReq.inputs) {
-      request.add_walletids(input.second);
-      signer.addSpender(std::make_shared<ScriptSpender>(input.first));
+      request.add_walletids(input.walletId_);
+      signer.addSpender(std::make_shared<ScriptSpender>(input.utxo_));
```
sergey-chernikov, line 462: We don't expose publicly accessible struct members with "_" suffix. So better to remove it.

WalletsLib/CoreWallet.cpp
```diff
@@ -1190,6 +1190,70 @@ BinaryData bs::core::SignMultiInputTX(const bs::core::wallet::TXMultiSignRequest
...
+
+   for (const auto &recipient : txMultiReq.recipients) {
+      signer.addRecipient(recipient);
+   }
+
+   signer.sign();
+
+   for (const auto &spender : spenders) {
+      const auto &itSig = inputSigs.find(spender.first);
+      if (itSig == inputSigs.end() || !spender.second->isSegWit()) {
```
sergey-chernikov, line 1245: Does this function work for SegWit inputs only? You have code above for legacy spenders init.

romaniy173: Yeah, if we are talking about mixing(more then one input type) - we use only this for segwit, so no segwit and legacy input mixing for now. If function about, it's about signing from one wallet - that function is possible for legacy

romaniy173: True, I will delete those function, this should be possible only for segwit

romaniy173: Done

BlocksettleNetworkingLib/HeadlessContainer.cpp
romaniy173: Done

PR #1899, opened 2020-04-23
Check index before lookup in vector

romaniy173:
Issue: fix crash when trying to get index which is out of bound


BlocksettleNetworkingLib/SelectedTransactionInputs.cpp
```diff
@@ -248,6 +248,10 @@ const UTXO& SelectedTransactionInputs::GetTransaction(size_t i) const
 
 bool SelectedTransactionInputs::IsTransactionSelected(size_t i) const
 {
+   if (i < 0 || i >= selection_.size()) {
```
pavel-kokolemin, line 251: no need to check if < 0 (`size_t` is unsigned)

romaniy173: True, fixed, thank you!

PR #1908, opened 2020-04-27
Add change address index argument to create tx request function

romaniy173:
Issue: I want to have possibility to send change address without change wallet in request to signer

Particular case: when signing legacy address input from hw device, we include 1 legacy wallet in tx(since we do not allow to missing legacy with segwit) and change address could be only segwit(we do not have possibility to create change legacy address). So we want to address itself and address index will defined on terminal side.

Note: I did not find any useful case to send change wallet in purpose within the inputs dependent  wallets in tx request while creating simple tx

### DIFF
--- a/BlocksettleNetworkingLib/TradesUtils.cpp
+++ b/BlocksettleNetworkingLib/TradesUtils.cpp
@@ -214,7 +214,7 @@ void bs::tradeutils::createPayin(bs::tradeutils::PayinArgs args, bs::tradeutils:
                         auto recipients = std::vector<std::shared_ptr<ScriptRecipient>>(1, recipient);
                         try {
                            result.signRequest = bs::sync::wallet::createTXRequest(args.inputXbtWallets, selectedInputs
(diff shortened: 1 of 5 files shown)


pavel-kokolemin:
> I want to have possibility to send change address without change wallet in request to signer
I think that's a bad idea because then signer will not be able check that it's genuine change address.

romaniy173:
> > I want to have possibility to send change address without change wallet in request to signer
> 
> I think that's a bad idea because then signer will not be able check that it's genuine change address (trough we don't have that right now but I'm working on it).
> 
> > Particular case: when signing legacy address input from hw device, we include 1 legacy wallet in tx(since we do not allow to missing legacy with segwit)
> 
> IMO proper fix would be to just include both wallets.

Good point, then I will close this PR and fix wallet detecting logic on Signer site instead. NP from my side, just thought that additional wallet which do not need in signer could bring issue, but now it's not make sense according to your comment, so I closing this PR

PR #1910, opened 2020-04-28
Bst 2675

pavel-kokolemin:
http://185.213.153.35:8081/browse/BST-2675

### DIFF
--- a/BlocksettleNetworkingLib/ProtobufHeadlessUtils.cpp
+++ b/BlocksettleNetworkingLib/ProtobufHeadlessUtils.cpp
@@ -15,6 +15,12 @@
 #include "CheckRecipSigner.h"
 #include "CoreHDLeaf.h"
 
+namespace {
+
+   const int kValidPathLength = 2;
+
+}
+
 headless::SignTxRequest bs::signer::coreTxRequestToPb(const bs::core::wallet::TXSignRequest &txSignReq
    , bool keepDuplicatedRecipients)
 {
@@ -89,7 +95,9 @@ bs::core::wallet::TXSignRequest pbTxRequestToCoreImpl(const headless::SignTxRequ
    }
    for (const auto &inputIndex : request.input_indices()) {
       if (!inputIndex.empty()) {
-         bs::hd::Path::fromString(inputIndex);
+         if (bs::hd::Path::fromString(inputIndex).length() != kValidPathLength) {
+            throw std::runtime_error("unexpected path length for UTXO input address");
+         }
       }
       txSignReq.inputIndices.push_back(inputIndex);
    }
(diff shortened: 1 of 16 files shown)


BlocksettleNetworkingLib/ProtobufHeadlessUtils.cpp
```diff
@@ -85,8 +87,12 @@ bs::core::wallet::TXSignRequest bs::signer::pbTxRequestToCore(const headless::Si
          txSignReq.inputs.push_back(utxo);
       }
    }
-   txSignReq.inputIndices.insert(txSignReq.inputIndices.end()
-      , request.input_indices().begin(), request.input_indices().end());
+   for (const auto &inputIndex : request.input_indices()) {
+      if (!inputIndex.empty()) {
+         bs::hd::Path::fromString(inputIndex);
```
sergey-chernikov, line 92: Where does the return value of this function go to?

BlocksettleNetworkingLib/ProtobufHeadlessUtils.cpp
```diff
@@ -103,6 +109,7 @@ bs::core::wallet::TXSignRequest bs::signer::pbTxRequestToCore(const headless::Si
    }
 
    if (request.has_change()) {
+      bs::hd::Path::fromString(request.change().index());
```
sergey-chernikov, line 112: Same here - is it just dropped? Why at least the result is not checked for emptyness / valid number of path elements?

WalletsLib/CoreWallet.cpp
```diff
@@ -658,38 +658,6 @@ bool wallet::TXSignRequest::isSourceOfTx(const Tx &signedTx) const
    }
 }
 
-bool wallet::TXSignRequest::populateSupportingTx(std::shared_ptr<ArmoryConnection> connPtr)
```
sergey-chernikov, line 661: Didn't find a replacement for this removed code. Is it just vanished?

BlocksettleNetworkingLib/ProtobufHeadlessUtils.cpp
pavel-kokolemin: Sure, let's check path length.

BlocksettleNetworkingLib/ProtobufHeadlessUtils.cpp
pavel-kokolemin: Fixed.

WalletsLib/CoreWallet.cpp
pavel-kokolemin: Moved out earlier to `CreateTransactionDialog::CreateTransaction`

PR #1912, opened 2020-04-28
BST-2678

pavel-kokolemin:
http://185.213.153.35:8081/browse/BST-2678
CC UTXO selection fixes (and remove duplicated code)
The problem was that we were selecting UTXOs without covering fee for CC part and this check sometimes failed:
```
   if (inputAmount < (spendVal + fee)) {
      throw std::overflow_error("Not enough inputs (" + std::to_string(inputAmount)
         + ") to spend " + std::to_string(spendVal + fee));
   }
```
No change in unit tests passing


BlocksettleNetworkingLib/Wallets/SyncWalletsManager.cpp
```diff
@@ -1944,15 +1955,22 @@ bs::core::wallet::TXSignRequest WalletsManager::createPartialTXRequest(uint64_t
          }
       }
 
+      // TODO: Remove this, looks like it affects only TestCCSettlement.InvalidChangeAmount
```
sergey-chernikov, line 1958: Actually it will avoid throwing in any case when no outputs are attached (which is still a valid case)

PR #1913, opened 2020-04-29
Add wallet purpose to otc offer

romaniy173:
Issue: we need to have wallet purpose when create payout for hw wallet


pavel-kokolemin:
I got a "path unknown" warning with trezor emulator testing OTC:

romaniy173:
> I got a "path unknown" warning with trezor emulator testing OTC:
> 

This is okay, because we send nested address and change as native address, you should to apply also fix from terminal - https://github.com/BlockSettle/common/pull/1913 to avoid this. In that fix we will get change address the same like input address, so for nested we will generate nested change address and this error should gone. Anyway I do not see this issue with real device. Maybe if it fail with terminal fix we need to investigate more

PR #1916, opened 2020-05-01
Fix bs::signer::coreTxRequestToPb regression

pavel-kokolemin:


pavel-kokolemin:
Already fixed in https://github.com/BlockSettle/common/pull/1915

PR #1920, opened 2020-05-03
Introduce new enc key for hw wallet

romaniy173:
Issue: there is no possible to distinguish hw wallet from hw offline wallet type. Introduce new encryption structure  HardwareEncKey which will be contain all hw wallet related data
http://185.213.153.35:8081/browse/BST-2681

### DIFF
--- a/BlocksettleNetworkingLib/HeadlessContainerListener.h
+++ b/BlocksettleNetworkingLib/HeadlessContainerListener.h
@@ -146,6 +146,7 @@ private:
    bool onSignSettlementPayoutTxRequest(const std::string &clientId
       , const Blocksettle::Communication::headless::RequestPacket &packet);
    bool onSignAuthAddrRevokeRequest(const std::string &clientId, const Blocksettle::Communication::headless::RequestPacket &);
+   bool onResolvePubSpenders(const std::string &clientId, const Blocksettle::Communication::headless::RequestPacket &packet);
    bool onCreateHDLeaf(const std::string &clientId, Blocksettle::Communication::headless::RequestPacket &packet);
    bool onPromoteHDWallet(const std::string& clientId, Blocksettle::Communication::headless::RequestPacket& packet);
    bool onSetUserId(const std::string &clientId, Blocksettle::Communication::headless::RequestPacket &packet);
--- a/BlocksettleNetworkingLib/HeadlessContainerListener.h
+++ b/BlocksettleNetworkingLib/HeadlessContainerListener.h
@@ -208,8 +208,6 @@ private:
 
    void sendSyncWallets(std::string clientId = {});
 
-
-
 private:
    std::shared_ptr<spdlog::logger>     logger_;
    ServerConnection                    *connection_{};
--- a/BlocksettleNetworkingLib/HeadlessContainerListener.h
+++ b/BlocksettleNetworkingLib/HeadlessContainerListener.h
@@ -206,6 +206,9 @@ private:
    void sendUpdateStatuses(std::string clientId = {});
 
    void sendSyncWallets(std::string clientId = {});
+
+
+
 private:
    std::shared_ptr<spdlog::logger>     logger_;
    ServerConnection                    *connection_{};
--- a/BlocksettleNetworkingLib/Wallets/SyncHDWallet.cpp
+++ b/BlocksettleNetworkingLib/Wallets/SyncHDWallet.cpp
@@ -33,8 +33,7 @@ hd::Wallet::Wallet(const bs::sync::WalletInfo &info, WalletSignerContainer *cont
    netType_ = getNetworkType();  // netType_ = info.netType ???
    bool isHw = false;
    for (auto type : info.encryptionTypes) {
-      if (bs::wallet::EncryptionType::Hardware == type ||
-         bs::wallet::EncryptionType::HardwareOffline == type) {
+      if (bs::wallet::EncryptionType::Hardware == type) {
          isHw = true;
          break;
       }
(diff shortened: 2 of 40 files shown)


sergey-chernikov:
Can you please clarify, why can't we combine hardware encryption type and isOffline flag? Why should we introduce a new encryption type for their combination (which looks senseless to me)?

romaniy173:
> Can you please clarify, why can't we combine hardware encryption type and isOffline flag? Why should we introduce a new encryption type for their combination (which looks senseless to me)?

Sorry for late response. The issue here that the hw wallet is offline and hardware in the first place, which assuming we could sign it only with hw device. Also right now in terminal we have each leaf separated for hw wallet(native, nested, legacy) and we do not allow mixing inputs of different type of leafs. Now after we export HW wallet as WO and import that file, we will have the usual WO wallet, which will allow us to mixing  different types inputs which we trying to avoid. 

Second point here - it's unclear what you mean by isOffline flag because looks like WO and offline wallet have the same meaning. If that so, combination Hardware encryption + isOffline(WO) - will give us possibility to sign tx with hw device, which we do not want for a WO wallet exported from out app. Thanks

sergey-chernikov:
What's the difference between the Hardware and HardwareOffline types then? Both require HW to sign them, and both don't allow mixing of inputs.

romaniy173:
> What's the difference between the Hardware and HardwareOffline types then? Both require HW to sign them, and both don't allow mixing of inputs.

Well, no, Hardware require device immediately to sign it and HardwareOffline will create a file .bin which you could sign later like with usual WO signing and it do not need any identification in a first place. But true that they both don't allow inputs mixing.

sergey-chernikov:
So, you're telling me, that exactly HardwareOffline = Hardware + WO, so why do we need to introduce a new encryption type? I still don't get it. If you need a signing modification behavior, introduce it with WO/Offline flag, not by new encryption type (which is actually not a new one).

BlocksettleNetworkingLib/HeadlessContainerListener.h
```diff
@@ -207,6 +207,9 @@ class HeadlessContainerListener : public ServerConnectionListener
    void sendUpdateStatuses(std::string clientId = {});
 
    void sendSyncWallets(std::string clientId = {});
+
+
+
```
sergey-chernikov, line 212: Can you not add so much empty space? One line is enough here.

BlocksettleNetworkingLib/Wallets/SyncHDWallet.cpp
```diff
@@ -447,16 +453,45 @@ void hd::Wallet::getSettlementPayinAddress(const SecureBinaryData &settlementID
...
+
+std::unique_ptr<bs::wallet::HardwareEncKey> bs::sync::hd::Wallet::getHwEncKey() const
+{
+   if (encryptionTypes_.empty() || encryptionKeys_.empty() ||
+      encryptionTypes_[0] != bs::wallet::EncryptionType::Hardware) {
+      return nullptr;
+   }
+
+   return std::unique_ptr<bs::wallet::HardwareEncKey>(
+      new bs::wallet::HardwareEncKey(encryptionKeys_[0]));
```
sergey-chernikov, line 490: Better to use `make_unique` (without std namespace) instead of `new`. BTW, why do you need allocation here? Can't it be a plain stack object (and an exception in case of failure to init it)?

WalletsLib/WalletEncryption.cpp
```diff
@@ -32,3 +32,38 @@ BinaryData mergeKeys(const BinaryData &a, const BinaryData &b)
...
+{
+}
+
+bs::wallet::HardwareEncKey::HardwareEncKey(BinaryData binaryData)
+{
+   if (sizeof(uint32_t) >= binaryData.getSize())
+      std::logic_error("Incorrect binary data size");
+
+   walletType_ = static_cast<WalletType>(
+      BinaryData::StrToIntLE<uint32_t>(binaryData.getSliceRef(0, sizeof(uint32_t))));
```
sergey-chernikov, line 48: It's much more convenient to use `BinaryReader` instead of direct access to BinaryData methods. But it's up to you.

WalletsLib/WalletEncryption.h
```diff
@@ -45,6 +45,27 @@ namespace bs {
          BinaryData        salt;
          SecureBinaryData  controlPassword;
       };
+
+      struct HardwareEncKey {
+         enum WalletType : uint32_t
+         {
+            Offline,
```
sergey-chernikov, line 52: I thought `Offline` is a flag to complement both Trezor and Ledger (i.e. Trezor + Offline and Ledger + Offline). If this isn't so - ok, it was just my impression then.

BlocksettleNetworkingLib/HeadlessContainerListener.h
romaniy173: Done, thanks

BlocksettleNetworkingLib/Wallets/SyncHDWallet.cpp
romaniy173: Done. And I do not like solution with exception, since this is normal situation when we trying to take key from non hw wallet.This function is private and using with above functions where we treat it correctly. Thanks

WalletsLib/WalletEncryption.cpp
romaniy173: Change to BinaryReader, just miss some function in first place, not obvious. Thanks

WalletsLib/WalletEncryption.h
romaniy173: Indeed there is no reason to treat hw offline wallet differently per vendor. Thanks

PR #1922, opened 2020-05-04
Fix TXSignRequest operators

pavel-kokolemin:
Can't create any transactions and OTC tests fail


sergey-chernikov:
Another similar solution is submitted in https://github.com/BlockSettle/common/pull/1919 already.

PR #1928, opened 2020-05-08
Move hw wallet sign functionality inside hd root wallet

romaniy173:
Issue: we want to hind hw sign functionality in hw root wallet to make possible to reuse it in different place

### DIFF
--- a/WalletsLib/CoreHDWallet.cpp
+++ b/WalletsLib/CoreHDWallet.cpp
@@ -13,6 +13,7 @@
 #include "SystemFileUtils.h"
 #include "Wallets.h"
 #include "Assets.h"
+#include "headless.pb.h"
 
 
 #define LOG(logger, method, ...) \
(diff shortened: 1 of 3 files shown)


WalletsLib/CoreHDWallet.cpp
```diff
@@ -345,6 +346,46 @@ void bs::core::hd::Wallet::createHwStructure(const bs::core::wallet::HwWalletInf
...
+   if (!leaf) {
+      throw std::logic_error(fmt::format("Cannot find wallet {}", request.walletIds.front()));
+   }
+
+   BinaryData signedTx;
+   if (isHardwareWallet()) {
+      bs::wallet::HardwareEncKey hwEncKey(encryptionKeys()[0]);
+
+      std::set<BinaryData> binaryData;
+      auto signedDeviceSigs = lbdPwdPrompts_.back()(binaryData);
```
sergey-chernikov, line 361: You need to check that `lbdPwdPrompts_` is not empty

romaniy173: Thank you, fixed

WalletsLib/CoreHDWallet.cpp
```diff
@@ -345,6 +346,50 @@ void bs::core::hd::Wallet::createHwStructure(const bs::core::wallet::HwWalletInf
...
+   if (!leaf) {
+      throw std::logic_error(fmt::format("Cannot find wallet {}", request.walletIds.front()));
+   }
+
+   BinaryData signedTx;
+   if (isHardwareWallet()) {
+      bs::wallet::HardwareEncKey hwEncKey(encryptionKeys()[0]);
+
+      if (lbdPwdPrompts_.empty()) {
+         throw std::logic_error(fmt::format("Incorrect hw wallet data {}", request.walletIds.front()));
```
sergey-chernikov, line 361: I suggest to use "password lambda not set" without walletId

PR #1939, opened 2020-05-15
Save userId in PB

pavel-kokolemin:


Blocksettle_proto/Communication/bs_server.proto
```diff
@@ -932,13 +933,36 @@ message CelerSetUserPropertiesRequest
    }
 }
 
+message CelerGetUserPropertiesRequest
+{
+   string email = 1;
+   string email_hash = 2;
+   string propery_name = 3;
```
sergey-chernikov, line 940: I see only one property here. Isn't it supposed to be repeated?

pavel-kokolemin: Fixed, thanks, copy/paste error.

PR #1940, opened 2020-05-16
Add bip39 support

romaniy173:
Issue: add bip39 support

### DIFF
--- a/CommonLib/Bip39.cpp
+++ b/CommonLib/Bip39.cpp
@@ -113,66 +113,52 @@ bool validate_mnemonic(const std::vector<std::string> &words, const std::vector<
    return std::equal(mnemonic.begin(), mnemonic.end(), words.begin());
 }
 
-bool pkcs5_pbkdf2(const uint8_t* passphrase, size_t passphrase_length,
-   const uint8_t* salt, size_t salt_length, uint8_t* key, size_t key_length,
-   size_t iterations)
+bool pkcs5_pbkdf2(const SecureBinaryData& passphrase, const BinaryData& salt,
+   SecureBinaryData& key, size_t iterations)
 {
-   uint8_t* asalt;
-   size_t asaltSize;
-   size_t count, index, iteration, length;
-   uint8_t buffer[64U];
-   uint8_t digest1[64U];
-   uint8_t digest2[64U];
+   BinaryData asalt = salt;
+   SecureBinaryData buffer(64U);
+   SecureBinaryData digest1(64U);
+   SecureBinaryData digest2(64U);
 
-   /* An iteration count of 0 is equivalent to a count of 1. */
-   /* A key_length of 0 is a no-op. */
-   /* A salt_length of 0 is perfectly valid. */
-
-   if (salt_length > SIZE_MAX - 4)
-      return false;
-   asaltSize = salt_length + 4;
-   asalt = new uint8_t[asaltSize];
-   if (asalt == nullptr)
-      return false;
-
-   memcpy(asalt, salt, salt_length);
-   for (count = 1; key_length > 0; count++)
+   size_t keyLength = key.getSize();
+   BinaryWriter packer(keyLength);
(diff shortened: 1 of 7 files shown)


CommonLib/Bip39.cpp
```diff
@@ -0,0 +1,178 @@
+/*
+
+***********************************************************************************
+* Copyright (C) 2016 - , BlockSettle AB
```
sergey-chernikov, line 4: Is it really copyrighted by BlockSettle, or you taken it from somewhere?

CommonLib/Bip39.cpp
```diff
@@ -0,0 +1,178 @@
...
+      key += length;
+      key_length -= length;
+   };
+
+   delete[] asalt;
+
+   return true;
+}
+
+SecureBinaryData pib39GetSeedFromMnemonic(std::string sentence)
```
sergey-chernikov, line 168: const ref is advised for arg

CommonLib/Bip39.cpp
```diff
@@ -0,0 +1,178 @@
...
+         }
+      }
+   }
+
+   chunk.resize(entropyBits / byteBits);
+   const auto mnemonic = create_mnemonic(chunk, dictionary);
+   return std::equal(mnemonic.begin(), mnemonic.end(), words.begin());
+}
+
+bool pkcs5_pbkdf2(const uint8_t* passphrase, size_t passphrase_length,
```
sergey-chernikov, line 116: passphrase and passphrase_length can be replaced by a single SecureBinaryData, same for salt/key - by BinaryData.

CommonLib/Bip39.h
```diff
@@ -0,0 +1,36 @@
...
+bool validate_mnemonic(const std::vector<std::string> &words,
+   const std::vector<std::vector<std::string>>& dictionaries);
+
+// pbkdf2 algorithm to get bip39 root seed without passphrase
+bool pkcs5_pbkdf2(const uint8_t* passphrase, size_t passphrase_length,
+   const uint8_t* salt, size_t salt_length, uint8_t* key, size_t key_length,
+   size_t iterations);
+
+// return bip39 root seed which could be converted to bip32 root key
+SecureBinaryData pib39GetSeedFromMnemonic(std::string sentence);
```
sergey-chernikov, line 34: What is "pib"? A typo?

build_scripts/bip_protocols_settings.py
```diff
@@ -0,0 +1,46 @@
...
+
+from component_configurator import Configurator
+
+
+class BipProtocolsSettings(Configurator):
+    def __init__(self, settings):
+        Configurator.__init__(self, settings)
+        ref_version = subprocess.check_output("git ls-remote https://github.com/bitcoin/bips.git master", shell=True)
+        self._version = ref_version.decode().split('\t')[0]
+        self._script_revision = '3'
```
sergey-chernikov, line 14: Why script version is already 3 for a newly added module?

CommonLib/Bip39.cpp
romaniy173: I have took it from other place, but since we do not sync 3rd party library and we use our internal hashing function/data structure, I do not think we should specify it by different copyright here, not sure here though.

CommonLib/Bip39.cpp
romaniy173: thanks, fixed

CommonLib/Bip39.cpp
romaniy173: change to use our internal data structure the whole function, thanks

CommonLib/Bip39.h
romaniy173: Typo, fixed. Thank you

build_scripts/bip_protocols_settings.py
romaniy173: My mistake, fixed. Thank you

PR #1947, opened 2020-05-20
Fix bip39 and add electrum seeds verification logic

romaniy173:
Issues:
1. Fix whitespaces in bip39 seeds detection
2. Add support for electrum wallet seeds verifying ( note : task with adding wallet from electrum seed is not done on this review since electrum derivation schema is different from one we using, so for now I only left seeds detection logic which is not using atm)

### DIFF
--- a/CommonLib/Bip39.cpp
+++ b/CommonLib/Bip39.cpp
@@ -53,8 +53,7 @@ namespace {
 
       size_t keyLength = key.getSize();
       BinaryWriter packer(keyLength);
-      for (size_t count = 1; keyLength > 0; count++)
-      {
+      for (size_t count = 1; keyLength > 0; count++) {
          asalt.append((count >> 24) & 0xff);
          asalt.append((count >> 16) & 0xff);
          asalt.append((count >> 8) & 0xff);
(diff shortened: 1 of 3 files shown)


CommonLib/Bip39.cpp
```diff
@@ -17,148 +17,246 @@
...
+      SecureBinaryData digest2(64U);
 
-   const size_t entropyBits = (entropy.getSize() * byteBits);
-   const size_t checkBits = (entropyBits / entropyBitDevisor);
-   const size_t totalBits = (entropyBits + checkBits);
-   const size_t wordCount = (totalBits / bitsPerMnemonicWord);
+      size_t keyLength = key.getSize();
+      BinaryWriter packer(keyLength);
+      for (size_t count = 1; keyLength > 0; count++)
+      {
```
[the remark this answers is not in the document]
romaniy173: Sure, thanks

CommonLib/Bip39.cpp
```diff
@@ -17,148 +17,242 @@
...
-         BtcUtils::getHMAC512(passphrase.getPtr(), passphrase.getSize(), digest1.getPtr(), digest1.getSize(),
-            digest2.getPtr());
-         digest1 = digest2;
-         for (size_t index = 0; index < buffer.getSize(); index++)
-            buffer[index] ^= digest1[index];
+   std::vector<std::string> words = splitMnemonicWords(sentence);
+   for (auto const &dictionary : dictionaries) {
+      if (validateBip39Mnemonic(words, dictionary)) {
+         return true;
+         break;
```
Ation, line 197: break after return?

romaniy173: fixed, thanks

PR #1983, opened 2020-06-09
Use XBTAmount to avoid rounding errors

sergey-chernikov:


WalletsLib/XBTAmount.h
```diff
@@ -18,35 +18,69 @@ namespace bs {
...
+      static BTCNumericTypes::satoshi_type convertFromBitcoinToSatoshi(BTCNumericTypes::balance_type amount);
+      static BTCNumericTypes::balance_type convertFromSatoshiToBitcoin(BTCNumericTypes::satoshi_type value);
+
+   private:
+      BTCNumericTypes::satoshi_type value_;
+   };
 
 }
 
+int64_t operator-(const bs::XBTAmount &a, const int64_t b);
```
Ation, line 82: Please remove arithmetic operators with int64_t. The whole idea of that class is to have explicit type

sergey-chernikov: XBTAmount doesn't negative values, which is sometimes required at subtraction.

pavel-kokolemin:
unit test added here - https://github.com/BlockSettle/terminal/pull/782

PR #1984, opened 2020-06-10
Add WebSocket connection

pavel-kokolemin:
BIP15X will be added later, TLS will be used for now.

### DIFF
--- a/BlocksettleNetworkingLib/ServerConnectionListener.h
+++ b/BlocksettleNetworkingLib/ServerConnectionListener.h
@@ -40,8 +40,8 @@ public:
    virtual void OnClientConnected(const std::string& clientId) = 0;
    virtual void OnClientDisconnected(const std::string& clientId) = 0;
 
-   virtual void OnPeerConnected(const std::string &) {}
-   virtual void OnPeerDisconnected(const std::string &) {}
+   virtual void OnPeerConnected(const std::string &ip) {}
+   virtual void OnPeerDisconnected(const std::string &ip) {}
 
    virtual void onClientError(const std::string &clientId, const std::string &error) {}
    virtual void onClientError(const std::string &clientId, ClientError errorCode, int socket) {}
--- a/BlocksettleNetworkingLib/WsConnection.h
+++ b/BlocksettleNetworkingLib/WsConnection.h
@@ -1,0 +1,55 @@
+/*
+
+***********************************************************************************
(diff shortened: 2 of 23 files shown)


BlocksettleNetworkingLib/ServerConnectionListener.h
```diff
@@ -40,8 +40,8 @@ class ServerConnectionListener
    virtual void OnClientConnected(const std::string& clientId) = 0;
    virtual void OnClientDisconnected(const std::string& clientId) = 0;
 
-   virtual void OnPeerConnected(const std::string &) {}
-   virtual void OnPeerDisconnected(const std::string &) {}
+   virtual void OnPeerConnected(const std::string &ip) {}
```
Ation, line 43: just one more warning?

BlocksettleNetworkingLib/WsConnection.h
```diff
@@ -0,0 +1,55 @@
...
+
+         size_t getSize() const
+         {
+            return data_.size() - kLwsPrePaddingSize;
+         }
+      };
+
+      extern const char *kProtocolNameWs;
+
+      const size_t kRxBufferSize = 16 * 1024;
```
Ation, line 46: this can be a constexpr

BlocksettleNetworkingLib/WsConnection.h
```diff
@@ -0,0 +1,55 @@
...
+
+#include <chrono>
+#include <cstdint>
+#include <string>
+#include <vector>
+
+namespace bs {
+   namespace network {
+
+      extern const size_t kLwsPrePaddingSize;
```
Ation, line 22: Why this variable exposed? It could be in source file as a constexpr.

BlocksettleNetworkingLib/ServerConnectionListener.h
pavel-kokolemin: why not to to use `-Wno-unused-variable` if that bothers you?

BlocksettleNetworkingLib/WsConnection.h
pavel-kokolemin: fixed, trough I don't see the point. `constexpr` are for guaranteed compile time computations which we don't use AFAIK.

BlocksettleNetworkingLib/WsConnection.h
pavel-kokolemin: sure, fixed

PR #1992, opened 2020-06-17
[FEATURE BRANCH] Auth address tiers

Ation:
**Do not merge till feature ready**


BlocksettleNetworkingLib/AuthAddressManager.cpp
```diff
@@ -91,8 +91,19 @@ bool AuthAddressManager::setup()
          return;
       }
       if (GetState(address) != state) {
-         logger_->info("Address verification {} for {}", to_string(state), address.display());
-         SetState(address, state);
+         logger_->info("Address verification on chain {} for {}", to_string(state), address.display());
+
+         if (state == AddressVerificationState::NotSubmitted) {
+            const auto &submittedAddresses = celerClient_->GetSubmittedAuthAddressSet();
```
goatpig, line 97: what is celer's involvement with auth addresses?

BlocksettleNetworkingLib/AuthAddressManager.cpp
```diff
@@ -91,8 +91,19 @@ bool AuthAddressManager::setup()
...
+            if (submittedAddresses.find(address.display()) != submittedAddresses.end()) {
+               SetState(address, AddressVerificationState::Submitted);
+            } else {
+               SetState(address, state);
```
goatpig, line 101: Looking at SetState, this call will return with no effect. Why call the method at all in a dedicated else branch? This suggests auth addresses that are presented are being tracked even when they're invalid. It's a bit confusing.

BlocksettleNetworkingLib/AuthAddressManager.cpp
Ation: we have list of "submitted auth addresses" saved as user property on celer. Since there is no way for terminal to know if address was submitted or not if terminal is clean setup and wallet was imported.

BlocksettleNetworkingLib/AuthAddressManager.cpp
Ation: state in this method - on-chain validation result.
SetState sort of auth address cache.

BlocksettleNetworkingLib/AuthAddressManager.cpp
goatpig: why is this done by celer and not the PB storage DB or genoa is what im asking

goatpig:
common submodule wasn't updated in PR to the PB repo btw.

BlocksettleNetworkingLib/AuthAddressManager.cpp
goatpig: Let me rephrase: the call to SetState at line 101 will have no effect. Why call it at all?

BlocksettleNetworkingLib/AuthAddressManager.cpp
Ation: because it was done 4 years ago

BlocksettleNetworkingLib/AuthAddressManager.cpp
Ation: It will change address state from InProgress to NotSubmitted

PR #1995, opened 2020-06-18
Support for RFQ automation

sergey-chernikov:


WalletsLib/Address.cpp
```diff
@@ -291,6 +291,9 @@ uint64_t bs::Address::getFeeForMaxVal(const std::vector<UTXO> &utxos, size_t txO
 
 std::string bs::Address::display() const
 {
+   if (empty()) {
+      return "<empty>";
```
Ation, line 295: Why we need this change? Maybe it is more "natural" to return "" for empty address. Since if we expect string it is obvious to check if that string have something as a sign that serialization went well.

sergey-chernikov: This is discussible. In some cases it's better to see the address is empty (so next deser from this string will throw an exception). I can change this to empty string ofc, just not sure this is the right solution.

Ation: What is a better tell that address is empty then an empty string?

sergey-chernikov: return value changed

PR #2005, opened 2020-07-01
Fix ui auth dialog

romaniy173:
Issue: http://185.213.153.35:8081/browse/BST-2769
1. Save trade settings in auth manager
2. Lookup for not only verified but also submitted addresses while create rfq


BlocksettleNetworkingLib/AuthAddressManager.h
```diff
@@ -194,6 +198,7 @@ private slots:
 
    std::shared_ptr<SignContainer>      signingContainer_;
    std::unordered_set<unsigned int>    signIdsRevoke_;
+   std::shared_ptr<bs::TradeSettings>  tradeSettings_{nullptr};
```
sergey-chernikov, line 201: No need to init shared_ptr with nullptr - its default constructor does the same

romaniy173: Just habit, fixed, thank you

PR #2008, opened 2020-07-01
Auth address validation refactor

sergey-chernikov:


goatpig:
pushed fixes, all tests should pass now

PR #2010, opened 2020-07-02
CC tracker server updates

pavel-kokolemin:


BlocksettleNetworkingLib/ColoredCoinServer.cpp
```diff
@@ -536,9 +537,10 @@ void CcTrackerClient::processParseCcCandidateTx(const bs::tracker_server::Respon
 
 
 CcTrackerServer::CcTrackerServer(const std::shared_ptr<spdlog::logger> &logger
-   , const std::shared_ptr<ArmoryConnection> &armory)
+   , const std::shared_ptr<ArmoryConnection> &armory, ServerConnection *server)
```
sergey-chernikov, line 540: Is there a reason you pass plain pointer here instead of the smart one? Sorry, can't see the usage.

pavel-kokolemin: Sure, let's use shared_ptr

PR #2011, opened 2020-07-07
Add shared MDInfo class

romaniy173:
Issue: add helpful class MDInfo in order to reduce creating similar classes in other places

### DIFF
--- a/BlocksettleNetworkingLib/CommonTypes.cpp
+++ b/BlocksettleNetworkingLib/CommonTypes.cpp
@@ -102,6 +102,15 @@ MDField MDField::get(const MDFields &fields, MDField::Type type) {
    return { MDField::Unknown, 0, QString() };
 }
 
+bs::network::MDInfo bs::network::MDField::get(const MDFields &fields)
+{
+   MDInfo mdInfo;
+   mdInfo.bidPrice = std::max(.0, bs::network::MDField::get(fields, bs::network::MDField::PriceBid).value);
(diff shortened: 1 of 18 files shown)


sergey-chernikov:
This change is not related to auth addresses - please rebase to bs_dev - we can then merge bs_dev to auth_address_tiers

BlocksettleNetworkingLib/CommonTypes.cpp
```diff
@@ -102,6 +102,15 @@ MDField MDField::get(const MDFields &fields, MDField::Type type) {
    return { MDField::Unknown, 0, QString() };
 }
 
+bs::network::MDInfo bs::network::MDField::get(const MDFields &fields)
+{
+   MDInfo mdInfo;
+   mdInfo.bidPrice = std::max(.0, bs::network::MDField::get(fields, bs::network::MDField::PriceBid).value);
+   mdInfo.askPrice = std::max(.0, bs::network::MDField::get(fields, bs::network::MDField::PriceOffer).value);
+   mdInfo.lastPrice = std::max(.0, bs::network::MDField::get(fields, bs::network::MDField::PriceLast).value);
```
sergey-chernikov, line 110: What's the point in std::max? If value is zero or positive, just return it. If it's negative, it's either intended (remember WTI options prices not long ago) or an error, and shouldn't be replaced by 0 in any case.

romaniy173: true, changed. Thank you

PR #2018, opened 2020-07-14
[WIP] WebSocket updates

pavel-kokolemin:
- Add session resuming to WsDataConnection/WsServerConnection.
  Needed when Cloudflare servers restarted.
  Reimplement something that we had with ZMQ (transparent TCP restarts)
- Add Bip*Connection that could handle BIP encryption for other connections.
  Superfluous TransportBIP15x* could be removed later.
- Fix BIP15X rekeying regression from client side.
- Remove adding Transport to ZMQ streams.

### DIFF
--- a/BlocksettleNetworkingLib/HeadlessContainer.cpp
+++ b/BlocksettleNetworkingLib/HeadlessContainer.cpp
@@ -11,13 +11,14 @@
 #include "HeadlessContainer.h"
 
 #include "BSErrorCodeStrings.h"
+#include "Bip15xDataConnection.h"
 #include "ConnectionManager.h"
 #include "DataConnection.h"
 #include "ProtobufHeadlessUtils.h"
 #include "SystemFileUtils.h"
 #include "Wallets/SyncHDWallet.h"
 #include "Wallets/SyncWalletsManager.h"
-#include "ZmqDataConnection.h"
+#include "WsDataConnection.h"
 
 #include <QCoreApplication>
 #include <QDataStream>
@@ -1338,13 +1339,6 @@ bool RemoteSigner::Start()
       return true;
    }
 
-   if (opMode() == OpMode::RemoteInproc) {
-      const auto zmqConn = std::dynamic_pointer_cast<ZmqDataConnection>(connection_);
-      if (zmqConn) {
-         zmqConn->SetZMQTransport(ZMQTransport::InprocTransport);
-      }
-   }
-
    {
       std::lock_guard<std::mutex> lock(mutex_);
       listener_ = std::make_shared<HeadlessListener>(logger_, connection_, netType_);
@@ -1445,9 +1439,8 @@ void RemoteSigner::RecreateConnection()
    try {
       bip15xTransport_ = std::make_shared<bs::network::TransportBIP15xClient>(logger_, params);
       bip15xTransport_->setKeyCb(cbNewKey_);
-      bip15xTransport_->setLocalHeartbeatInterval();
-      auto conn = std::make_shared<ZmqBinaryConnection>(logger_, bip15xTransport_);
-      conn->SetContext(connectionManager_->zmqContext());
+      auto wsConn = std::make_unique<WsDataConnection>(logger_, WsDataConnectionParams{});
+      auto conn = std::make_shared<Bip15xDataConnection>(logger_, std::move(wsConn), bip15xTransport_);
       connection_ = std::move(conn);
 
       headlessConnFinished_ = false;
--- a/BlocksettleNetworkingLib/TransportBIP15xServer.cpp
+++ b/BlocksettleNetworkingLib/TransportBIP15xServer.cpp
@@ -147,6 +147,9 @@ TransportBIP15xServer::~TransportBIP15xServer() noexcept
 
 void TransportBIP15xServer::rekey(const std::string &clientId)
 {
+   //XXX : rekey disabled
+   return;
+
    auto connection = GetConnection(clientId);
    if (connection == nullptr) {
       logger_->error("[TransportBIP15xServer::rekey] can't find connection for {}", BinaryData::fromString(clientId).toHexStr());
--- a/BlocksettleNetworkingLib/TransportBIP15xServer.cpp
+++ b/BlocksettleNetworkingLib/TransportBIP15xServer.cpp
@@ -151,11 +151,6 @@ TransportBIP15xServer::~TransportBIP15xServer() noexcept
    }
 }
 
-void TransportBIP15xServer::periodicCheck()
-{
-   checkHeartbeats();
-}
-
 void TransportBIP15xServer::rekey(const std::string &clientId)
 {
    auto connection = GetConnection(clientId);
@@ -216,112 +211,69 @@ std::unique_ptr<BIP15xPeer> TransportBIP15xServer::getClientKey(const std::strin
 // OUTPUT: None
 // RETURN: None
 void TransportBIP15xServer::processIncomingData(const std::string &encData
-   , const std::string &clientID, int socket)
+   , const std::string &clientID)
 {
-   const auto &connData = setBIP151Connection(clientID);
-   if (!connData || !connData->isValid) {
-      logger_->error("[TransportBIP15xServer::processIncomingData] disconnected/invalid"
-         " client {} connection sent {} bytes", encData.size(), bs::toHex(clientID), encData.size());
-      if (clientErrorCb_) {
-         clientErrorCb_(clientID, "missing connection data");
+   const auto &connData = GetConnection(clientID);
+   assert(connData);
+   if (!connData->isValid) {
+      return;
+      logger_->debug("[TransportBIP15xServer::processIncomingData] disconnected/invalid"
+         " client {} connection sent {} bytes", bs::toHex(clientID), encData.size(), encData.size());
+      return;
+   }
+
+   auto payload = BinaryData::fromString(encData);
+
+   // Decrypt only if the BIP 151 handshake is complete.
+   if (connData->bip151HandshakeCompleted_) {
+      //decrypt packet
+      auto result = connData->encData_->decryptPacket(
+         payload.getPtr(), payload.getSize(),
+         payload.getPtr(), payload.getSize());
+
+      if (result != 0) {
+         logger_->error("[TransportBIP15xServer::processIncomingData] packet"
+            " {} [{} bytes] decryption failed: {}", payload.toHexStr(), payload.getSize(), result);
+         connData->isValid = false;
+         clientErrorCb_(clientID, ServerConnectionListener::ClientError::HandshakeFailed, connData->details);
+         return;
+      }
+      payload.resize(payload.getSize() - POLY1305MACLEN);
+   }
+
+   // Deserialize packet.
+   const auto &msg = bip15x::Message::parse(payload);
+   if (!msg.isValid()) {
+      if (logger_) {
+         logger_->error("[TransportBIP15xServer::processIncomingData] deserialization failed");
+      }
+      connData->isValid = false;
+      clientErrorCb_(clientID, ServerConnectionListener::ClientError::HandshakeFailed, connData->details);
+      return;
+   }
+
+   if (msg.getType() > bip15x::MsgType::AEAD_Threshold) {
+      if (!processAEADHandshake(msg, clientID)) {
+         reportFatalError(connData);
+         return;
       }
       return;
    }
 
-   const auto &packets = bip15x::MessageBuilder::parsePackets(BinaryData::fromString(
-      accumulBuf_.empty() ? encData : accumulBuf_ + encData));
-   if (packets.empty()) {
-      logger_->error("[TransportBIP15xServer::processIncomingData] packet deser failed");
-      if (clientErrorCb_) {
-         clientErrorCb_(clientID, "deserialization failed");
-      }
+   // We can now safely obtain the full message.
+   const BinaryDataRef &outMsg = msg.getData();
+
+   // We shouldn't get here but just in case....
+   if (connData->encData_->getBIP150State() != BIP150State::SUCCESS) {
+      logger_->error("[TransportBIP15xServer::processIncomingData] encryption"
+         " handshake is incomplete");
+      reportFatalError(connData);
       return;
    }
-   if (packets.at(0).empty()) {  // special condition to accumulate more data
-      accumulBuf_.append(encData);
-      return;
-   }
-   accumulBuf_.clear();
 
-   for (auto payload : packets) {
-      // Decrypt only if the BIP 151 handshake is complete.
-      if (connData->bip151HandshakeCompleted_) {
-         //decrypt packet
-         auto result = connData->encData_->decryptPacket(
-            payload.getPtr(), payload.getSize(),
-            payload.getPtr(), payload.getSize());
-
-         if (result != 0) {
-            logger_->error("[TransportBIP15xServer::processIncomingData] packet"
-               " {} [{} bytes] decryption failed: {}", payload.toHexStr(), payload.getSize(), result);
-            if (clientErrorCb_) {
-               clientErrorCb_(clientID, "packet decryption failed");
-            }
-            connData->isValid = false;
-            return;
-         }
-         payload.resize(payload.getSize() - POLY1305MACLEN);
-      }
-
-      // Deserialize packet.
-      const auto &msg = bip15x::Message::parse(payload);
-      if (!msg.isValid()) {
-         if (logger_) {
-            logger_->error("[TransportBIP15xServer::processIncomingData] deserialization failed");
-         }
-         if (clientErrorCb_) {
-            clientErrorCb_(clientID, "deserialization failed");
-         }
-         return;
-      }
-
-      // If we're still handshaking, take the next step. (No fragments allowed.)
-      if (msg.getType() == bip15x::MsgType::Heartbeat) {
-         UpdateClientHeartbeatTimestamp(clientID);
-
-         const auto &packet = bip15x::MessageBuilder(bip15x::MsgType::Heartbeat)
-            .encryptIfNeeded(connData->encData_.get()).build();
-         if (sendDataCb_) {
-            sendDataCb_(clientID, packet.toBinStr());
-         }
-         return;
-      }
-
-      if (msg.getType() > bip15x::MsgType::AEAD_Threshold) {
-         if (!processAEADHandshake(msg, clientID)) {
-            if (logger_) {
-               logger_->error("[TransportBIP15xServer::processIncomingData] handshake failed");
-            }
-            if (clientErrorCb_) {
-               clientErrorCb_(clientID, "handshake failed");
-            }
-            if (clientSocketErrorCb_) {
-               clientSocketErrorCb_(clientID, ServerConnectionListener::HandshakeFailed, socket);
-            }
-            return;
-         }
-         continue;
-      }
-
-      // We can now safely obtain the full message.
-      const BinaryDataRef &outMsg = msg.getData();
-
-      // We shouldn't get here but just in case....
-      if (connData->encData_->getBIP150State() != BIP150State::SUCCESS) {
-         if (logger_) {
-            logger_->error("[TransportBIP15xServer::processIncomingData] encryption"
-               " handshake is incomplete");
-         }
-         if (clientErrorCb_) {
-            clientErrorCb_(clientID, "encryption handshake failure");
-         }
-         return;
-      }
-
-      // Pass the final data up the chain.
-      if (dataReceivedCb_) {
-         dataReceivedCb_(clientID, outMsg.toBinStr());
-      }
+   // Pass the final data up the chain.
+   if (dataReceivedCb_) {
+      dataReceivedCb_(clientID, outMsg.toBinStr());
    }
 }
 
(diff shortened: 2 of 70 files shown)


sergey-chernikov:
Some general notes on the PR family:
1. BIP is a Bitcoin Improvement Proposal, IMO better to precise file and class naming, 15x suffix seems appropriate
2. You use generic transport and connection in e.g. BipConnection (it can be ZMQ + SSL at some point) - why not name it something like CombinedTransportConnection or just TransportConnection?
3. We need working tests (TestNetwork in the terminal repo) to be sure all edge cases are covered properly.

pavel-kokolemin:
>     1. BIP is a Bitcoin Improvement Proposal, IMO better to precise file and class naming, 15x suffix seems appropriate
Sure, fixed. I just don't like that it's too verbose.
>     2. You use generic transport and connection in e.g. BipConnection (it can be ZMQ + SSL at some point) - why not name it something like CombinedTransportConnection or just TransportConnection?
My plan is to remove TransportConnection (merge it with `Bip15x*Connection`) someday, so there will be only `Bip15x*Connection` that will only work above `Ws*Connection`. SSL will be handled by `WsDataConnection` only. We won't be using ZMQ in new code. So I guess we could leave it as is.

>     3. We need working tests (TestNetwork in the terminal repo) to be sure all edge cases are covered properly.
I'm working on that. I don't expect any problems here, so I guess this could be merged to unblock others (as staging is broken now).

pavel-kokolemin:
Network unit tests reverted, except:
- `ZMQ_BIP15X_MalformedData` is not relevant now and removed
- `ZMQ_BIP15X_ConnectionTimeout` will be fixed a bit later

sergey-chernikov:
2. The idea behind connection and transport separation was to encapsulate all connection instances into base type and don't care about BIP15x specifics when just sending/receiving data. All logic was put to transport. If you gonna merge it back, you'll have to declare BIP15x-specific connection types again everywhere

BlocksettleNetworkingLib/WsDataConnection.cpp
```diff
@@ -135,9 +150,9 @@ int WsDataConnection::callback(lws *wsi, int reason, void *user, void *in, size_
 {
    switch (reason) {
       case LWS_CALLBACK_OPENSSL_LOAD_EXTRA_CLIENT_VERIFY_CERTS: {
-//         if (!params_.useSsl || params_.caBundleSize == 0) {
+         if (!params_.useSsl || params_.caBundleSize == 0) {
```
sergey-chernikov, line 153: Why do we need SSL re-enabled? It should be superseded by BIP15x

BlocksettleNetworkingLib/WsServerConnection.cpp
```diff
@@ -108,7 +108,11 @@ void WsServerConnection::stopServer()
       return;
    }
 
-   stopped_ = true;
+   DataToSend toSend{kAllClientsId, WsPacket::close()};
+   {  std::lock_guard<std::mutex> lock(mutex_);
```
sergey-chernikov, line 112: Please put `{` in block and code line to separate rows

BlocksettleNetworkingLib/WsServerConnection.cpp
```diff
@@ -118,146 +122,378 @@ void WsServerConnection::stopServer()
...
-            return -1;
-         }
-         if (rc != static_cast<int>(packet.getSize())) {
-             SPDLOG_LOGGER_ERROR(logger_, "write truncated");
-             return -1;
-         }
-
-         if (!client.packets.empty()) {
-            lws_callback_on_writable(client.wsi);
+         auto &connection = connections_.at(wsi);
```
sergey-chernikov, line 314: at() may throw under some conditions - do you catch elsewhere?

BlocksettleNetworkingLib/WsServerConnection.cpp
```diff
@@ -118,146 +122,378 @@ void WsServerConnection::stopServer()
...
 }
 
-std::thread::id WsServerConnection::listenThreadId() const
+bool WsServerConnection::done() const
 {
-   return listenThread_.get_id();
+   if (!shuttingDown_) {
+      return false;
+   }
+   {  std::lock_guard<std::mutex> lock(mutex_);
```
sergey-chernikov, line 420: Same here for `{` and new line

BlocksettleNetworkingLib/WsServerConnection.cpp
```diff
@@ -118,146 +122,378 @@ void WsServerConnection::stopServer()
...
-   // So we use multicast here.
-   size_t cntClients = clients_.size();
-   for (const auto &item : clients_) {
-      if (SendDataToClient(item.first, data)) {
-         cntClients--;
-      }
-   }
-   return (cntClients == 0);
+{
+   return SendDataToClient(kAllClientsId, data);
```
sergey-chernikov, line 496: Did you read the deleted comment? Did you test the broadcast of encrypted data?

BlocksettleNetworkingLib/ZmqServerConnection.cpp
```diff
@@ -452,30 +367,26 @@ bool ZmqServerConnection::QueueDataToSend(const std::string& clientId, const std
...
 {
-   if (transport_) {
-      transport_->periodicCheck();
-   }
 }
 
 void ZmqServerConnection::SendDataToDataSocket()
 {
-   decltype(dataQueue_) pendingData;
+   std::deque<DataToSend> pendingData;
```
sergey-chernikov, line 374: Why this change? From my perspective the deleted line is more obvious and clear, that it's a destination for swap of `dataQueue_`.

BlocksettleNetworkingLib/ZmqServerConnection.cpp
```diff
@@ -452,30 +367,26 @@ bool ZmqServerConnection::QueueDataToSend(const std::string& clientId, const std
...
+
    {
       FastLock locker{dataQueueLock_};
       pendingData.swap(dataQueue_);
    }
 
    for (const auto &dataPacket : pendingData) {
-      int result = zmq_send(dataSocket_.get(), dataPacket.clientId.c_str()
-         , dataPacket.clientId.size(), ZMQ_SNDMORE);
+      int result = zmq_send(dataSocket_.get(), dataPacket.clientId.c_str(), dataPacket.clientId.size(), ZMQ_SNDMORE);
```
sergey-chernikov, line 382: What this and next change actually does? Except making this line longer and scrollable on small displays.

BlocksettleNetworkingLib/ZmqStreamServerConnection.cpp
```diff
@@ -9,18 +9,17 @@
...
-#include <zmq.h>
-#include <spdlog/spdlog.h>
+
 #include "ActiveStreamClient.h"
 #include "FastLock.h"
 #include "MessageHolder.h"
-#include "Transport.h"
 
+#include <zmq.h>
+#include <spdlog/spdlog.h>
```
sergey-chernikov, line 18: Again - is this a blind rollback of everything? Regardless of useful changes?

BlocksettleNetworkingLib/HeadlessContainer.cpp
```diff
@@ -1338,13 +1339,6 @@ bool RemoteSigner::Start()
       return true;
    }
 
-   if (opMode() == OpMode::RemoteInproc) {
```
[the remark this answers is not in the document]
pavel-kokolemin: Everything works fine and `RemoteInproc` was not used, `LocalInproc` should be enough.
> I mean, any kind of inproc connection (even not a localhost one)?
Sorry, I don't understand the question. Why would you need `RemoteInproc` type? If you want in-process signer then local connection is the sane type to use.

BlocksettleNetworkingLib/TransportBIP15xServer.cpp
```diff
@@ -216,112 +211,69 @@ std::unique_ptr<BIP15xPeer> TransportBIP15xServer::getClientKey(const std::strin
...
+   auto payload = BinaryData::fromString(encData);
 
-      // If we're still handshaking, take the next step. (No fragments allowed.)
-      if (msg.getType() == bip15x::MsgType::Heartbeat) {
-         UpdateClientHeartbeatTimestamp(clientID);
+   // Decrypt only if the BIP 151 handshake is complete.
+   if (connData->bip151HandshakeCompleted_) {
+      //decrypt packet
+      auto result = connData->encData_->decryptPacket(
+         payload.getPtr(), payload.getSize(),
```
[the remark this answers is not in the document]
pavel-kokolemin: `Ws*Connection` does not use fragmented packets.

BlocksettleNetworkingLib/TransportBIP15xServer.cpp
```diff
@@ -216,112 +211,69 @@ std::unique_ptr<BIP15xPeer> TransportBIP15xServer::getClientKey(const std::strin
...
-            logger_->error("[TransportBIP15xServer::processIncomingData] encryption"
-               " handshake is incomplete");
-         }
-         if (clientErrorCb_) {
-            clientErrorCb_(clientID, "encryption handshake failure");
-         }
+   if (msg.getType() > bip15x::MsgType::AEAD_Threshold) {
+      if (!processAEADHandshake(msg, clientID)) {
+         reportFatalError(connData);
          return;
```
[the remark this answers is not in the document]
pavel-kokolemin: I don't know, it was already there, just checked old `ZMQ_BIP15X_ServerConnection.cpp`.

BlocksettleNetworkingLib/WsDataConnection.cpp
pavel-kokolemin: `WsDataConnection` could be used without BIP15x.

BlocksettleNetworkingLib/WsServerConnection.cpp
pavel-kokolemin: Sure, updated.

BlocksettleNetworkingLib/WsServerConnection.cpp
pavel-kokolemin: `at` here is on purpose for sanity check

BlocksettleNetworkingLib/WsServerConnection.cpp
pavel-kokolemin: Sure, fixed.

BlocksettleNetworkingLib/WsServerConnection.cpp
pavel-kokolemin: `WsServerConnection` does not broadcast encrypted data now, that is up to `Bip15xServerConnection` and it will encrypt it as needed.

BlocksettleNetworkingLib/ZmqServerConnection.cpp
pavel-kokolemin: Sure, reverted.

BlocksettleNetworkingLib/ZmqServerConnection.cpp
pavel-kokolemin: Just reverted old file.

BlocksettleNetworkingLib/ZmqStreamServerConnection.cpp
pavel-kokolemin: What exactly we are missing here? ZMQ will be deprecated and removed soon.

pavel-kokolemin:
> The idea behind connection and transport separation was to encapsulate all connection instances into base type and don't care about BIP15x specifics when just sending/receiving data.

Yes, I got that and still don't see the point.

> you'll have to declare BIP15x-specific connection types again everywhere

Sorry, I don't see why I would need to do that.

BlocksettleNetworkingLib/HeadlessContainer.cpp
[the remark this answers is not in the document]
sergey-chernikov: It's usable, but previously it was in-memory connection from ZMQ - is it possible to use it now?

BlocksettleNetworkingLib/TransportBIP15xServer.cpp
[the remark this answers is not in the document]
sergey-chernikov: Did you notice ZMQ in my comment? I'm talking generally, not only WS

BlocksettleNetworkingLib/TransportBIP15xServer.cpp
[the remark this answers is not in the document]
sergey-chernikov: Ofc, but you added new return below - better to remove one, that is unnecessary.

BlocksettleNetworkingLib/WsDataConnection.cpp
sergey-chernikov: It could, but should it?

BlocksettleNetworkingLib/WsServerConnection.cpp
sergey-chernikov: `at` can throw an exception - is it caught somewhere? Or is it ok to crash the whole process?

BlocksettleNetworkingLib/WsServerConnection.cpp
sergey-chernikov: So, you're saying that Bip15xServerConnection can *broadcast* data to multiple clients, encrypted once? Encrypted by what client's key?

BlocksettleNetworkingLib/ZmqStreamServerConnection.cpp
sergey-chernikov: Just the changes that are not necessary to revert. Headers reorder in this case.

sergey-chernikov:
> Yes, I got that and still don't see the point.

OOD, encapsulation. Other classes need to know about the connection as less as possible. If you declare a BIP15x-specific connection and start to use its specific methods, no other replacement for it will be possible to use without changing all classes that use it.

BlocksettleNetworkingLib/WsDataConnection.cpp
pavel-kokolemin: I think yes, where we don't need to bother with all that Bip15x nonsense. Whole world uses SSL while we trying to reinvent wheel.

BlocksettleNetworkingLib/WsServerConnection.cpp
pavel-kokolemin: Bip15xServerConnection will send encrypted packets to separate clients. And this is tested in unit tests.

pavel-kokolemin:
> > Yes, I got that and still don't see the point.
> 
> OOD, encapsulation. Other classes need to know about the connection as less as possible. If you declare a BIP15x-specific connection and start to use its specific methods, no other replacement for it will be possible to use without changing all classes that use it.

We already have `DataConnection` and `ServerConnection` as abstraction. And so nobody uses BIP15x-specific stuff unless strictly needed.

BlocksettleNetworkingLib/WsDataConnection.cpp
sergey-chernikov: Strong statement regarding BIP15x nonsense...
@goatpig please decide here

BlocksettleNetworkingLib/WsServerConnection.cpp
pavel-kokolemin: Better to crash whole process in case of logic errors than to have security breach as result of UB exploitation. So yes, much better.

BlocksettleNetworkingLib/WsServerConnection.cpp
sergey-chernikov: Then this shouldn't be called *broadcast*. What's the real broadcast for in WsServerConnection then?

BlocksettleNetworkingLib/TransportBIP15xServer.cpp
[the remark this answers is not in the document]
pavel-kokolemin: Yes, sure, fixed.

sergey-chernikov:
> We already have `DataConnection` and `ServerConnection` as abstraction. And so nobody uses BIP15x-specific stuff unless strictly needed.

Yes, this is what I was trying to achieve. But in case you obsolete this state of things, you will start using specific BIP15x methods (e.g. to manage keys) in a connection instead of the transport.

BlocksettleNetworkingLib/TransportBIP15xServer.cpp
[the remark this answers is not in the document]
pavel-kokolemin: Yes, I saw it, but we will not ZMQ with BIP15X.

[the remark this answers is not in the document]
sergey-chernikov: What if it will be just plain TCP socket? Should we re-implement the transport?

BlocksettleNetworkingLib/HeadlessContainer.cpp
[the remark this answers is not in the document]
pavel-kokolemin: No, we don't have in-memory connection right now. I guess using loopback address should work here.

pavel-kokolemin:
> you will start using specific BIP15x methods (e.g. to manage keys) in a connection instead of the transport.

Yes, sure, but I don't see the problem. If you need to manage keys you will need to do it somewhere. Right now it is encapsulated in `Bip15x*Connection`. If you don't need to manage keys you use abstract `DataConnection`/`ServerConnection`. If we ever need to you to use BIP15x over ZMQ again, we will use `Bip15x*Connection` with `Zmq*Connection` and `Zmq*Connection` will know nothing about BIP15x. And there is won't be need to add bunch of methods and new classes here and there like we just had.

BlocksettleNetworkingLib/WsServerConnection.cpp
sergey-chernikov: I wasn't asking if it's better - only if it's ok or not. Not sure though if we can't do it more gracefully - but ok, let's keep it as is.

BlocksettleNetworkingLib/WsServerConnection.cpp
pavel-kokolemin: `WsServerConnection` does not know and should not know about any of that, if it got command to send packets to all clients it will do so. `Bip15xServerConnection` should do what is needed here (encrypt packets separately).

BlocksettleNetworkingLib/TransportBIP15xServer.cpp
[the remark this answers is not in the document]
sergey-chernikov: Not yet, I guess

BlocksettleNetworkingLib/TransportBIP15xServer.cpp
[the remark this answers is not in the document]
pavel-kokolemin: Right, perhaps in such case using Transport would be a good idea to hide some of that `BIP15x` code. But I really doubt that is a good idea. We won't be able use Cloudflare for DDoS protection, nginx for routing etc.

BlocksettleNetworkingLib/WsDataConnection.cpp
pavel-kokolemin: We spend too much time here trying to solve already solved problem. And implementing home brew encryption is very rarely a good idea. For bitcoin BIP15X might have sense as it's decentralized. We don't need any of that.

pavel-kokolemin: For example we have had a bug when memory was beyond heap in `BIP151Session::decPayload`. That was accidental detected by ASAN. It's direct way for security exploitation. That bug was fixed but who knows what else we had here.

BlocksettleNetworkingLib/WsServerConnection.cpp
sergey-chernikov: The question here - will it do so properly? You have a special receiver to maintain broadcast - is it compatible with SendDataToAllClients in Bip15ServerConnection? And if it's not used - what's the use case for kAllClientsId at all?

BlocksettleNetworkingLib/ZmqServerConnection.cpp
```diff
@@ -230,24 +164,21 @@ void ZmqServerConnection::listenFunction()
...
-   poll_items[ZmqServerConnection::MonitorSocketIndex].events = ZMQ_POLLIN;
-
    logger_->debug("[{}] poll thread started for {}", __func__, connectionName_);
 
    int errorCount = 0;
 
    while (true) {
       int periodMs = std::chrono::duration_cast<std::chrono::milliseconds>(kHearthbeatCheckPeriod).count();
-      int result = zmq_poll(poll_items, 3, periodMs);
+      int result = zmq_poll(poll_items, sizeof(poll_items) / sizeof(poll_items[0]), periodMs);
```
Ation, line 181: You have compile time variable. Make it constexpr, don't use 30 years old C tricks.

pavel-kokolemin: Sorry, I don't understand what are you referring to. How `constexpr` will help here?

BlocksettleNetworkingLib/WsServerConnection.cpp
pavel-kokolemin: > The question here - will it do so properly?
It will do so properly as far as it concerns `WsServerConnection`. Not sure what are we discussing here, unit tests cover this.

pavel-kokolemin:
PR redirected to `auth_address_tiers` branch - https://github.com/BlockSettle/common/pull/2022

PR #2022, opened 2020-07-20
Websocket reconnects

pavel-kokolemin:


BlocksettleNetworkingLib/PublisherServerConnection.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+   void OnDataFromClient(const std::string& clientId, const std::string& data) override
+   {
+   }
+
+   void OnClientConnected(const std::string &clientId, const Details &details) override
+   {
+      std::lock_guard<std::mutex> lock(owner_->mutex_);
+      owner_->conn_->SendDataToClient(clientId, owner_->welcomeMsg_);
+      owner_->conn_->SendDataToClient(clientId, owner_->latestDataMsg_);
```
Ation, line 26: Why do we need to send latest message? Welcome message exists exactly for that

BlocksettleNetworkingLib/ZmqServerConnection.cpp
```diff
@@ -230,24 +164,21 @@ void ZmqServerConnection::listenFunction()
...
-   poll_items[ZmqServerConnection::MonitorSocketIndex].events = ZMQ_POLLIN;
-
    logger_->debug("[{}] poll thread started for {}", __func__, connectionName_);
 
    int errorCount = 0;
 
    while (true) {
       int periodMs = std::chrono::duration_cast<std::chrono::milliseconds>(kHearthbeatCheckPeriod).count();
-      int result = zmq_poll(poll_items, 3, periodMs);
+      int result = zmq_poll(poll_items, sizeof(poll_items) / sizeof(poll_items[0]), periodMs);
```
Ation, line 181: please use constexpr. This is compile time variable

BlocksettleNetworkingLib/PublisherServerConnection.cpp
pavel-kokolemin: Right, fixed, thanks

BlocksettleNetworkingLib/ZmqServerConnection.cpp
pavel-kokolemin: Sure, should be fixed now. Trough I'm 99% sure this is already optimized by compiler.

PR #2023, opened 2020-07-20
AuthAddressLogic: AWA shutdown and recovery

goatpig:
fix AuthAddressLogic shutdown sequence

### DIFF
--- a/BlocksettleNetworkingLib/AuthAddressLogic.cpp
+++ b/BlocksettleNetworkingLib/AuthAddressLogic.cpp
@@ -311,7 +311,9 @@ void AuthAddressValidator::waitOnRefresh(const std::string& id)
          break;
       }
       catch (const ArmoryThreading::StackTimedOutException &) {
-         break;
+         if (stopped_) {
+            break;
+         }
       }
    }
 }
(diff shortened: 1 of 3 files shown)


BlocksettleNetworkingLib/AuthAddressLogic.cpp
```diff
@@ -289,14 +302,17 @@ void AuthAddressValidator::waitOnRefresh(const std::string& id)
...
+         auto&& notifId = refreshQueue_.pop_front(std::chrono::seconds{ 5 });
+         if ((notifId == idRef) || stopped_) {
             break;
          }
       }
       catch (const ArmoryThreading::StopBlockingLoop&) {
          break;
       }
+      catch (const ArmoryThreading::StackTimedOutException &) {
+         break;
```
goatpig, line 314: This will kill the wait in 5 if pop_front times out, it should be continue, not break

sergey-chernikov: True, made it conditional

goatpig: The common side changes work, need to conclude the conversation on the PB test side.

goatpig: This is ready to go

PR #2025, opened 2020-07-21
Fix AuthLogic shutdown race

sergey-chernikov:


BlocksettleNetworkingLib/AuthAddressLogic.cpp
```diff
@@ -1135,6 +1213,73 @@ bool AuthAddressLogic::isValid(
...
+            return;
+         }
+
+         if (!pathState.revokedPaths_.empty()) {
+            //validation output was spent by user
+            cb(addr, AddressVerificationState::Revoked);
+            return;
+         }
+         //address has history and no validation outputs
+         cb(addr, AddressVerificationState::Tainted);
```
goatpig, line 1274: this should return after firing the callback

BlocksettleNetworkingLib/AuthAddressLogic.cpp
```diff
@@ -450,26 +448,53 @@ bool AuthAddressValidator::goOnline(const ResultCb &cb)
...
-   }
+   //grab all txouts
+   lambdas_->getOutpointsForAddresses(addrVec, cb, topBlock_, zcIndex_);
+}
 
+unsigned AuthAddressValidator::update(const OutpointBatch &batch)
+{
+   std::unique_lock<std::mutex> lock(updateMutex_);
+   if (!lambdas_ || stopped_) {
+      std::cout << "stopped 1\n";
```
goatpig, line 488: left over print statement

BlocksettleNetworkingLib/AuthAddressLogic.cpp
```diff
@@ -450,26 +448,53 @@ bool AuthAddressValidator::goOnline(const ResultCb &cb)
...
+      return UINT32_MAX;
+   }
    unsigned opCount = 0;
-   for (auto& outpointPair : batch.outpoints_) {
+   for (const auto &outpointPair : batch.outpoints_) {
       if (stopped_) {
-         return 0;
+         std::cout << "stopped 2\n";
```
goatpig, line 494: same as above

BlocksettleNetworkingLib/AuthAddressLogic.cpp
sergey-chernikov: Thank you, fixed

PR #2027, opened 2020-07-23
Add workaround for failed assert on Windows

pavel-kokolemin:


build_scripts/websockets_settings.py
```diff
@@ -101,6 +102,16 @@ def config(self):
         return result == 0
 
     def make_windows(self):
+        # Workaroung for failed assert on Windows
```
Ation, line 105: Since it is for assert only, can we limit patching for debug build only?

pavel-kokolemin: yes, sure, done

PR #2029, opened 2020-07-24
Stop auth address state notifications if they are not new

Ation:


BlocksettleNetworkingLib/AuthAddressManager.cpp
```diff
@@ -557,10 +557,12 @@ void AuthAddressManager::SetValidationState(const bs::Address &addr, AddressVeri
 
    SetExplicitState(addr, mappedState);
 
-   if (mappedState == AuthAddressState::Verified) {
+   if (mappedState == AuthAddressState::Verified
```
goatpig, line 560: This won't emit a notification for addresses that go from submitted to verified

PR #2044, opened 2020-08-11
Offline sign ccgen auth list

Ation:


Ation:
Starting manual testing on staging

PR #2045, opened 2020-08-13
Ssl connection

pavel-kokolemin:


BlocksettleNetworkingLib/SslDataConnection.cpp
```diff
@@ -0,0 +1,303 @@
...
+}
+
+bool SslDataConnection::openConnection(const std::string &host, const std::string &port
+   , DataConnectionListener *listener)
+{
+   closeConnection();
+
+   listener_ = listener;
+   host_ = host;
+   port_ = std::stoi(port);
```
sergey-chernikov, line 62: Not critical issue though, but we should support port resolution from /etc/services at some point.

PR #2047, opened 2020-08-13
Force TSLv1.3 in WsDataConnection

pavel-kokolemin:


BlocksettleNetworkingLib/SslDataConnection.cpp
```diff
@@ -70,7 +70,8 @@ bool SslDataConnection::openConnection(const std::string &host, const std::strin
    info.gid = -1;
    info.uid = -1;
    info.retry_and_idle_policy = bs::network::ws::defaultRetryAndIdlePolicy();
-   info.options = params_.useSsl ? LWS_SERVER_OPTION_DO_SSL_GLOBAL_INIT : 0;
+   info.options = 0;
+   info.options |= params_.useSsl ? LWS_SERVER_OPTION_DO_SSL_GLOBAL_INIT : 0;
```
Ation, line 74: I don't get the difference

pavel-kokolemin: Just using same format as in `SslDataConnection` to make sure both files are updated.

Ation: I mean that 
a = 0
a |= X;
is identical to
a = X;

Ation: so it just looks wrong. it should be either preserving existing flags or set new value. no need to set to 0 first

pavel-kokolemin: sure, fixed

PR #2078, opened 2020-09-03
[WIP] Allow using static XBT XINERAMA

pavel-kokolemin:


Ation:
I'm not really sure what problem is fixed here?

pavel-kokolemin:
> I'm not really sure what problem is fixed here?

Just trying to fix some linking problem. Looks like it did not work, need test it more, sorry.

PR #2079, opened 2020-09-03
Fix ws SSL error when mutliple context registered

pavel-kokolemin:


BlocksettleNetworkingLib/SslDataConnection.cpp
```diff
@@ -43,6 +43,8 @@ SslDataConnection::SslDataConnection(const std::shared_ptr<spdlog::logger> &logg
    assert(params_.useSsl || !params_.verifyCallback);
    assert(params_.useSsl || params_.cert.empty());
    assert(params_.useSsl || params_.privKey.empty());
+
+   ws::globalInit(params_.useSsl);
```
sergey-chernikov, line 47: what if we use both SSL and non-SSL connections in the same process?

pavel-kokolemin: It will work fine, `ws::globalInit` will do what needs to be done depending on passed flag.
I tested this with terminal.

sergey-chernikov: I'm sure it is, my question is more about the init sequence of a single global state (and possible race conditions here). I.e. if we first init an SSL connection and then a non-SSL one, will the SSL connection be in the proper state after that (same situation, but vice versa). I guess we need some test around it.

pavel-kokolemin: There is no race as we use `once_flag`, so I don't see any problems here. For non-SSL connection `ws::globalInit` has no effect.

sergey-chernikov (approved):

PR #2087, opened 2020-09-07
CC fixes

pavel-kokolemin:


BlocksettleNetworkingLib/Wallets/SyncWalletsManager.cpp
```diff
@@ -1923,29 +1924,34 @@ bs::core::wallet::TXSignRequest WalletsManager::createPartialTXRequest(uint64_t
...
          for (const auto &recipient : recipients.second) {
-            txSize += recipient->getSize();
+            baseSize += recipient->getSize();
          }
       }
-      auto weight = 4 * txSize + witnessSize;
-      prevPartTxSize = (weight + 3) / 4;
-   }
+      // CC output
+      auto rec = std::make_shared<Recipient_P2WPKH>(CryptoPRNG::generateRandom(20), 1000);
```
Ation, line 1947: It is always 31. Why do you need to call a PRNG for that?

WalletsLib/CheckRecipSigner.cpp
```diff
@@ -417,3 +417,22 @@ NetworkType getNetworkType()
    default:       return NetworkType::RegTest;
    }
 }
+
+uint64_t bs::estimateVSize(const Signer &signer)
```
Ation, line 421: If this method used only once in tests it should be there.

BlocksettleNetworkingLib/Wallets/SyncWalletsManager.cpp
pavel-kokolemin: sure, fixed

WalletsLib/CheckRecipSigner.cpp
pavel-kokolemin: it's used on server too

Ation (approved):

PR #2094, opened 2020-09-16
VS2019 build fixes

sergey-chernikov:


build_scripts/qt_settings.py
```diff
@@ -90,16 +90,16 @@ def config(self):
...
         command.append('-I{}'.format(os.path.join(self.openssl.get_install_dir(),'include')))
 
-        if self._project_settings.on_osx():
-            command.append('-L/usr/local/opt/mysql@5.7/lib')
-            command.append('-I/usr/local/opt/mysql@5.7/include')
-            command.append('-I/usr/local/opt/mysql@5.7/include/mysql')
+#        if self._project_settings.on_osx():
+#            command.append('-L/usr/local/opt/mysql@5.7/lib')
+#            command.append('-I/usr/local/opt/mysql@5.7/include')
+#            command.append('-I/usr/local/opt/mysql@5.7/include/mysql')
```
Ation, line 103: Please check server build

Ation (changes requested):
Please update script revisions for changed settings scripts ( except Qt )

build_scripts/gtest_settings.py
```diff
@@ -69,7 +70,9 @@ def make_windows(self):
         print('Start building GTest')
         print(' '.join(command))
 
-        result = subprocess.call(command)
+        #result = subprocess.call(command)
```
Ation, line 73: why do we need that change?
what it have to do with visual studio?

sergey-chernikov: It's something wrong with CMake on Windows - it can't properly parse double quotes (") if passing an arg as a list. Works fine with string for some reason. I'll make it platform-conditional in all other cases (like in websocket_settings atm).

build_scripts/qt_settings.py
sergey-chernikov: Server build now works - see https://github.com/scomil/BS_Private/pull/1256

Ation (approved):

PR #2095, opened 2020-09-17
Fix shutdown crash in RetryingDataConnection

pavel-kokolemin:


BlocksettleNetworkingLib/RetryingDataConnection.cpp
```diff
@@ -57,6 +57,7 @@ RetryingDataConnection::RetryingDataConnection(const std::shared_ptr<spdlog::log
 
 RetryingDataConnection::~RetryingDataConnection()
 {
+   shutingDown_ = true;
```
sergey-chernikov, line 60: Typo - should be "shutting"

pavel-kokolemin: fixed, thanks

sergey-chernikov (approved):

PR #2121, opened 2020-12-14
Allow close clients from SslServerConnection

pavel-kokolemin:


Ation:
Already done on my branch

PR #2123, opened 2020-12-15
Api server 2

pavel-kokolemin:


Ation (approved):

BlocksettleNetworkingLib/SslServerConnection.cpp
```diff
@@ -188,10 +191,8 @@ int SslServerConnection::callback(lws *wsi, int reason, void *user, void *in, si
             if (clientIt != clients_.end()) {
                SPDLOG_LOGGER_DEBUG(logger_, "force close client {}", bs::toHex(clientId));
                auto clientWsi = clientIt->second.wsi;
-               if (clientWsi) {
-                  lws_close_reason(clientWsi, LWS_CLOSE_STATUS_PROTOCOL_ERR, nullptr, 0);
-                  lws_set_timeout(clientWsi, PENDING_TIMEOUT_USER_OK, LWS_TO_KILL_SYNC);
-               }
+               lws_close_reason(clientWsi, LWS_CLOSE_STATUS_PROTOCOL_ERR, nullptr, 0);
+               lws_set_timeout(clientWsi, PENDING_TIMEOUT_USER_OK, kForcedDisconnectTimeSeconds);
```
Ation, line 195: https://github.com/warmcat/libwebsockets/issues/1463

Ation (changes requested):
Previous implementation looks correct. Actually it is how it was done in the first place in WsServerConnection

BlocksettleNetworkingLib/SslServerConnection.cpp
pavel-kokolemin: if we use `LWS_TO_KILL_SYNC` here, server will not send pending packets, for example with failed login error message

pavel-kokolemin:
> Previous implementation looks correct. Actually it is how it was done in the first place in WsServerConnection

WsServerConnection is broken too, it won't send pending packets.

PR #2129, opened 2020-12-23
Add deliverable type

pavel-kokolemin:


Blocksettle_proto/Communication/bs_api_server.proto
```diff
@@ -26,6 +26,7 @@ message Request
         uint64 xbt_amount = 1;
         double user_expected_price = 2;
         bs.types.Side side = 3;
+        string product_name = 4;
```
Ation, line 29: product name (XBT/EUR1) is not the same as product type (1day_deliverable). please fix naming for variables, it might lead to bugs caused by misreading the code.

Ation (changes requested):

Blocksettle_proto/Communication/bs_api_server.proto
pavel-kokolemin: should be fixed now

PR #2141, opened 2021-01-15
Workaround for failing checksum check

sergey-chernikov:


pavel-kokolemin:
This does not help, I got same crash:

```
Failed to start signer:unexpected seed string
```

I will check myself why creating new wallet works correctly but import fails.

pavel-kokolemin:
it works now, thanks
what about paper backups?
this only fixes the problem with digital backups, same happens with paper backups too

sergey-chernikov:
Added the same "fix" as for digital backups.

pavel-kokolemin (approved):

PR #2145, opened 2021-01-18
Terminal arch refactor

sergey-chernikov:
Lots of changes to support shared bus adapters for Armory, wallets and on-chain trackers.


BlocksettleNetworkingLib/ApplicationSettings.cpp
```diff
@@ -180,7 +178,7 @@ ApplicationSettings::ApplicationSettings(const QString &appName
       { AutoSigning,             SettingDef(QLatin1String("AutoSigning"), false) },
       { ExtConnName,             SettingDef(QLatin1String("ExtConnName")) },
       { ExtConnHost,             SettingDef(QLatin1String("ExtConnHost")) },
-      { ExtConnPort,             SettingDef(QLatin1String("ExtConnPort")) },
+      { ExtConnPort,             SettingDef(QLatin1String("ExtConnPort"), 4567) },
```
goatpig, line 181: what's this hardcoded port about?

BlocksettleNetworkingLib/ArmoryServersProvider.cpp
```diff
@@ -153,12 +153,12 @@ int ArmoryServersProvider::indexOfIpPort(const std::string &srvIPPort) const
    return -1;
 }
 
-int ArmoryServersProvider::getIndexOfMainNetServer() const
+int ArmoryServersProvider::getIndexOfMainNetServer()
```
goatpig, line 156: why was the const removed?

BlocksettleNetworkingLib/ApplicationSettings.cpp
sergey-chernikov: not hardcoded - it's a default value (if not specified in terminal .ini file)
This is now used as an API port for external terminal connections. Previously was used for connecting the terminal to some external JSON API, which was used in AQ scripts.

BlocksettleNetworkingLib/Adapters/OnChainTrackerAdapter.cpp
```diff
@@ -0,0 +1,393 @@
...
+   const ArmoryMessage_WalletRegistered& response)
+{
+   authVerificator_->pushRefreshID({ BinaryData::fromString(std::to_string(msgId)) });
+   return true;
+}
+
+bool OnChainTrackerAdapter::processOutpoints(uint64_t msgId
+   , const ArmoryMessage_OutpointsForAddrList& response)
+{
+   const auto& it = outpointCallbacks_.find(msgId);
```
goatpig, line 301: I dont understand how the callback can be registered after armory returns with the outpoint data

BlocksettleNetworkingLib/Adapters/SignerClient.h
```diff
@@ -0,0 +1,188 @@
...
+   SignerClient(const std::shared_ptr<spdlog::logger> &
+      , const std::shared_ptr<bs::message::User> &);
+   ~SignerClient() override = default;
+
+   void Start() override {}
+   void Connect() override {}
+
+   [[deprecated]] bs::signer::RequestId signTXRequest(const bs::core::wallet::TXSignRequest &
+      , TXSignMode mode = TXSignMode::Full, bool keepDuplicatedRecipients = false) override { return 0; }
+   void signTXRequest(const bs::core::wallet::TXSignRequest&
```
goatpig, line 52: I gather this class is for WO management but shouldn't the sign methods throw instead of noop? It comes to reason that a signer client meant for WO management should never have to handle a sign request.

BlocksettleNetworkingLib/ArmoryObject.cpp
```diff
@@ -13,7 +13,7 @@
 #include <QFile>
 #include <QProcess>
 #include <QPointer>
-
+#include <QThread>
```
goatpig, line 16: how much Qt is there in common? GUI?

BlocksettleNetworkingLib/AssetManager.cpp
```diff
@@ -12,9 +12,9 @@
 #include <QMutexLocker>
 #include <spdlog/spdlog.h>
 #include "AssetManager.h"
-#include "CelerClient.h"
-#include "CelerFindSubledgersForAccountSequence.h"
-#include "CelerGetAssignedAccountsListSequence.h"
+#include "Celer/CelerClient.h"
```
goatpig, line 15: Also how much Celer is in common?

BlocksettleNetworkingLib/AuthAddressManager.cpp
```diff
@@ -147,10 +156,7 @@ AuthAddressManager::ReadyError AuthAddressManager::readyError() const
    if (!HaveBSAddressList()) {
       return ReadyError::MissingAddressList;
    }
-   if (!armory_) {
-      return ReadyError::MissingArmoryPtr;
-   }
-   if (!armory_->isOnline()) {
+   if (armory_ && !armory_->isOnline()) {
```
goatpig, line 159: This would return NoError if armory_ == nullptr. Is that the desired behavior?

BlocksettleNetworkingLib/AssetManager.cpp
sergey-chernikov: Not much atm. Also AssetManager will be obsoleted, too, in favor of AssetsAdapter.

BlocksettleNetworkingLib/HeadlessContainer.cpp
```diff
@@ -61,6 +61,40 @@ NetworkType HeadlessContainer::mapNetworkType(headless::NetworkType netType)
...
+   SPDLOG_LOGGER_INFO(logger_, "remote signer has been disconnected");
+   isConnected_ = false;
+   isReady_ = false;
+   tryEmitError(HeadlessContainer::SignerGoesOffline
+      , QObject::tr("Remote signer disconnected"));
+}
+
+void HeadlessListener::tryEmitError(SignContainer::ConnectionError errorCode, const QString &msg)
+{
+   // Try to send error only once because only first error should be relevant.
```
goatpig, line 75: Why though?

BlocksettleNetworkingLib/Adapters/SignerClient.h
sergey-chernikov: returning 0 RequestId means the same, but throwing is more indicative, ofc

BlocksettleNetworkingLib/ArmoryObject.cpp
sergey-chernikov: Qt GUI is now concentrated only in QtGUI plugin of the terminal. Qt code in common is some core stuff like strings, files, pointers, processes and threads, which will be gradually eliminated.

BlocksettleNetworkingLib/HeadlessContainer.cpp
sergey-chernikov: no idea to be honest - I just reformatted that stuff, AFAIR
could be related to GUI pop-ups to avoid their cascading on a single error

BlocksettleNetworkingLib/Adapters/OnChainTrackerAdapter.cpp
sergey-chernikov: this line retrieves the already registered callback by msgId of the outpoints response
or, maybe, I didn't get your question?

BlocksettleNetworkingLib/Wallets/SyncHDLeaf.cpp
```diff
@@ -645,125 +686,6 @@ void hd::Leaf::topUpAddressPool(bool extInt, const std::function<void()> &cb)
       walletId(), lookup, extInt, fillUpAddressPoolCallback);
 }
 
-void hd::Leaf::scan(const std::function<void(bs::sync::SyncState)> &cb)
```
goatpig, line 648: Where did this go?

BlocksettleNetworkingLib/AuthAddressManager.cpp
sergey-chernikov: Fixed, though AuthAddressManager will be obsoleted, too, in the end.

BlocksettleNetworkingLib/Wallets/SyncHDLeaf.cpp
sergey-chernikov: Now scanning is performed in WalletsAdapter

BlocksettleNetworkingLib/Adapters/OnChainTrackerAdapter.cpp
goatpig: I meant to reference line 303, where the comment suggests that failing to find the callback can be remedied by waiting a bit longer (it returns false, implying the message will be requeued for later processing). 

The only way the reprocessing of this message could succeed is if the callback for this msgid is registered at some point after the data has been returned from armory, which seems odd to me (the message would essentially permanently bloat the queue).

BlocksettleNetworkingLib/Adapters/SignerClient.h
goatpig: If you look at the header, all the void virtual methods are defined as noops. A throw there would be a lot better. Returning for those methods that actually return something is fine.

BlocksettleNetworkingLib/HeadlessContainer.cpp
goatpig: I see calls to tryEmitError are mostly preceded by a call to logger_->error. Ideally tryEmitError should do the error logging as a way to guarantee there is a trace of all submitted errors, even though some of them won't be escalated. This would clean up the code too.
This isn't in the scope of this PR, but making a note of it for a later change, or as a PR to this PR, would be desirable.

BlocksettleNetworkingLib/Wallets/SyncHDLeaf.h
```diff
@@ -91,17 +98,21 @@ namespace bs {
             const bs::hd::Path &path() const { return path_; }
             bs::hd::Path::Elem index() const { return static_cast<bs::hd::Path::Elem>(path_.get(-1)); }
             bs::hd::Purpose purpose() const;
+            bool extOnly() const { return isExtOnly_; }
 
-            std::vector<std::string> registerWallet(const std::shared_ptr<ArmoryConnection> &armory = nullptr
+            [[deprecated]] std::vector<std::string> registerWallet(const std::shared_ptr<ArmoryConnection> &armory = nullptr
```
goatpig, line 103: I'm guessing this was moved to WalletAdapter as well?

BlocksettleNetworkingLib/Wallets/SyncHDWallet.cpp
```diff
@@ -25,7 +25,7 @@ using namespace bs::sync;
 
 hd::Wallet::Wallet(const bs::sync::WalletInfo &info, WalletSignerContainer *container
    , const std::shared_ptr<spdlog::logger> &logger)
-   : walletId_(info.id), name_(info.name), desc_(info.description)
+   : walletId_(*info.ids.cbegin()), name_(info.name), desc_(info.description)
```
goatpig, line 28: what guarantee is there info.ids has at least one entry?

BlocksettleNetworkingLib/Adapters/OnChainTrackerAdapter.cpp
sergey-chernikov: ok, changed to abort with error

BlocksettleNetworkingLib/Wallets/SyncHDWallet.cpp
sergey-chernikov: there's no guarantee, but wallets should typically have at least one id
if ids is empty, it will crash here and will give us an indication that something went wrong on the upper level

BlocksettleNetworkingLib/Wallets/SyncHDLeaf.h
sergey-chernikov: yes, exactly
all wallet registrations are now handled in WalletsAdapter (and BlockchainAdapter)

BlocksettleNetworkingLib/Wallets/SyncHDWallet.cpp
```diff
@@ -94,16 +94,16 @@ void hd::Wallet::synchronize(const std::function<void()> &cbDone)
          for (const auto &leafData : grpData.leaves) {
             auto leaf = group->getLeaf(leafData.path);
             if (!leaf) {
-               leaf = group->createLeaf(leafData.path, leafData.id);
+               leaf = group->createLeaf(leafData.path, *leafData.ids.cbegin());
```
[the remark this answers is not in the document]
sergey-chernikov: same consideration here, though added a check (as here it's technically possible)

BlocksettleNetworkingLib/HeadlessContainer.cpp
sergey-chernikov: Moved all outside error logging into tryEmitError. As for the initial question, the check for single emission was introduced by @pavel-kokolemin, AFAIR (and it was way long ago) - better to ask him here.

BlocksettleNetworkingLib/Adapters/BlockchainAdapter.cpp
```diff
@@ -0,0 +1,1589 @@
...
+{
+   if (armoryPtr_) {
+      sendLoadingBC();
+      init(armoryPtr_.get());
+      feeEstimationsCache_ = std::make_shared<BitcoinFeeCache>(logger_, armoryPtr_);
+      onStateChanged(armoryPtr_->state());
+   }
+   else {
+      ArmoryMessage msg;
+      msg.mutable_settings_request();  // broadcast - ask for someone to provide settings
```
goatpig, line 143: what happens if multiple adapters return settings?

BlocksettleNetworkingLib/Adapters/BlockchainAdapter.cpp
```diff
@@ -0,0 +1,1589 @@
...
+         const bool result = futureObj.get();
+         if (!result) { // stop armory connection loop if server key was rejected
+            armory->needsBreakConnectionLoop_ = true;
+            armory->setState(ArmoryState::Cancelled);
+         }
+         connKeyProm_.reset();
+         return result;*/
+         return true;
+      });
+      armoryPtr_ = armory;
```
goatpig, line 233: This seems to reset the armoryPtr_ on reception of a settings message, irreverant of current state of armoryPtr_.

BlocksettleNetworkingLib/Adapters/BlockchainAdapter.cpp
```diff
@@ -0,0 +1,1589 @@
...
+   bs::message::Envelope env{ 0, user_, nullptr, {}, {}, msg.SerializeAsString() };
+   pushFill(env);
+}
+
+void BlockchainAdapter::onZCInvalidated(const std::set<BinaryData> &ids)
+{
+   ArmoryMessage msg;
+   auto zcInv = msg.mutable_zc_invalidated();
+   for (const auto& id : ids) {
+      zcInv->add_tx_hashes(id.toBinStr());   //TODO: maybe add only pending TXs?
```
goatpig, line 379: //TODO: maybe add only pending TXs?

This db notification covers all ZC affecting the registered wallets, i.e. it is for ZC that made it previously in the mempool and were then ejected (possibly not marked as pending anymore within adapters)

BlocksettleNetworkingLib/Adapters/BlockchainAdapter.cpp
```diff
@@ -0,0 +1,1589 @@
...
+{
+   bs::Address address;
+   try {
+      address = std::move(bs::Address::fromAddressString(addrStr));
+   }
+   catch (const std::exception& e) {
+      logger_->error("[{}] invalid address string: {}", __func__, e.what());
+      return true;
+   }
+   const auto& walletId = CryptoPRNG::generateRandom(8).toHexStr();
```
goatpig, line 1146: This looks like it is used to register a single address for monitoring. I'm guessing this is used for monitoring settlement addresses. Using a fortuna RNG here instead of the main PRNG would be preferable.

BlocksettleNetworkingLib/Adapters/BlockchainAdapter.cpp
sergey-chernikov: yes - reception of settings message typically means armory needs to connect/reconnect anew (and happens once on program start basically).

BlocksettleNetworkingLib/Adapters/BlockchainAdapter.cpp
sergey-chernikov: Armory will reconnect multiple times, in the order of settings responses handling. But this should not happen normally (I don't say "never").

BlocksettleNetworkingLib/Adapters/BlockchainAdapter.cpp
sergey-chernikov: fixed

BlocksettleNetworkingLib/Adapters/BlockchainAdapter.cpp
goatpig: I commented here to reply the TODO's question. It is possible (yet unlikely) that invalidated ZC notifications come for hashes that are not monitored. 

This mostly concerns the PB and payin/payout monitoring. It isn't nearly as relevant for the terminal, but since this code is now shared by both, I'd rather all invalidated hashes are propagated regardless of pending status. I dont think that BlockchainAdapter should implement any kind of filtering on db messages, it should be agnostic in that capacity.

DB message filtering should be controlled by the set of registered addresses as far as BlockchainAdapter is concerned. Higher level adapters that rely on BlockchainAdapter can implement their own local filtering, i.e. the terminal can ignore ZC notifications for hashes it doesn't track, but the PB has code to deal with those. BA shouldn't pick and choose who gets what notification.

BlocksettleNetworkingLib/Adapters/BlockchainAdapter.cpp
goatpig: Use BtcUtils::fortuna_.generateRandom(unsigned). 

PRNG_Fortuna needs to be persisted otherwise you're burning through seeds needlessly. Either use the static one in BtcUtils or add one as a member of BlockchainAdapter. You don't have anything to do besides that, it will seed/reseed on its own.

BlocksettleNetworkingLib/Adapters/WalletsAdapter.cpp
```diff
@@ -0,0 +1,2501 @@
...
+         loadWallet(wallet);
+      }
+   };
+   signerClient_->syncWalletInfo(cbWalletsReceived);
+}
+
+void WalletsAdapter::loadWallet(const bs::sync::WalletInfo &info)
+{
+   logger_->debug("[WalletsManager::syncWallets] syncing wallet {} ({} {})"
+      , *info.ids.cbegin(), info.name, info.description);
```
goatpig, line 163: I'm generally confused as to why bs::sync::WalletInfo smashes all ids into a single vector. Can't struct have more explicit members for the various ids it uses? What kind of ids does it carry in that vector actually?

BlocksettleNetworkingLib/Adapters/WalletsAdapter.cpp
```diff
@@ -0,0 +1,2501 @@
...
+         }
+      } catch (const std::exception &e) {
+         logger_->error("[WalletsAdapter::loadWallet] failed to create HD wallet "
+            "{}: {}", *info.ids.cbegin(), e.what());
+      }
+      break;
+   }
+
+   case bs::sync::WalletFormat::Settlement:
+      throw std::runtime_error("not supported");
```
goatpig, line 205: is this WIP or is the functionality carried in elsewhere?

BlocksettleNetworkingLib/Adapters/BlockchainAdapter.cpp
sergey-chernikov: The problem with BlockchainAdapter used in different projects is there's no dedicated settings adapter. You can see it in the terminal, but in PB it's PrivBridgeAdapter for example. In some other project it can be something else. So that's why I introduce this broadcast - BlockchainAdapter asks anyone to send connection settings when ready to accept them and then uses them to connect. If some adapter sends them unsolicitedly, there can be no BlockchainAdapter created at that moment.

BlocksettleNetworkingLib/Adapters/BlockchainAdapter.cpp
sergey-chernikov: ok, removed that comment, left everything as-is

BlocksettleNetworkingLib/Adapters/WalletsAdapter.cpp
sergey-chernikov: Well, it actually can incapsulate the following wallets info: root HD wallet (single wallet id), HD leaf (1 or 2 ids), plain wallet (like in Armory - 1 wallet id), or any other extensions to BIP32 wallets. So it made as much universal as possible intentionally.

BlocksettleNetworkingLib/Adapters/WalletsAdapter.cpp
sergey-chernikov: Separate settlement wallets (as a dedicated lmdb file) are obsoleted by auth leaf nested leaves, but wallet format constant is still there.

BlocksettleNetworkingLib/Adapters/WalletsAdapter.cpp
```diff
@@ -0,0 +1,2501 @@
...
+   for (const auto &leaf : wallet->getLeaves()) {
+      addWallet(leaf);
+   }
+}
+
+void WalletsAdapter::addWallet(const std::shared_ptr<Wallet> &wallet)
+{
+   auto ccLeaf = std::dynamic_pointer_cast<bs::sync::hd::CCLeaf>(wallet);
+   if (ccLeaf) {
+//      ccLeaf->setCCDataResolver(ccResolver_);
```
goatpig, line 381: Can you add a log entry that CC is currently disabled? Don't know how long it's gonna take until we fix this.

BlocksettleNetworkingLib/Adapters/WalletsAdapter.cpp
```diff
@@ -0,0 +1,2501 @@
...
+   } else {
+      wallets_[wallet->walletId()] = wallet;
+   }
+
+   if ((wallet->type() == bs::core::wallet::Type::Authentication)) {
+      authAddressWallet_ = wallet;
+      logger_->debug("[WalletsAdapter::addWallet] auth leaf {} created", wallet->walletId());
+      WalletsMessage msg;
+      msg.set_auth_leaf_created(authAddressWallet_->walletId());
+      Envelope env{ 0, ownUser_, nullptr, {}, {}, msg.SerializeAsString() };
```
goatpig, line 398: Why does auth leaf creation lead to a broadcast event? What listens on this?

BlocksettleNetworkingLib/Adapters/WalletsAdapter.cpp
```diff
@@ -0,0 +1,2501 @@
...
+}
+
+void WalletsAdapter::processWalletBal(const ArmoryMessage_WalletBalanceResponse &response)
+{
+   for (const auto &byWallet : response.balances()) {
+      auto &balanceData = walletBalances_[byWallet.wallet_id()];
+      balanceData.walletBalance.totalBalance = byWallet.full_balance();
+      balanceData.walletBalance.unconfirmedBalance = byWallet.unconfirmed_balance();
+//      balanceData.walletBalance.spendableBalance = byWallet.spendable_balance();
+      balanceData.walletBalance.spendableBalance = balanceData.walletBalance.totalBalance
```
goatpig, line 759: why is spendable balance set that way? scott shenanigans? do tests pass like this?

BlocksettleNetworkingLib/Adapters/WalletsAdapter.cpp
```diff
@@ -0,0 +1,2501 @@
...
+         sendWalletReady(wallet->walletId());
+      }
+      if (st != bs::sync::SyncState::Failure) {
+         scanWallet(wallet, isExt);
+      }
+   };
+   signerClient_->syncAddressBatch(wallet->walletId(), usedAndRegAddrs, cb);
+}
+
+void WalletsAdapter::sendTxNRequest(const std::string &walletId)
```
goatpig, line 859: TxN stands for Tx Number?

sergey-chernikov: yes

BlocksettleNetworkingLib/Adapters/WalletsAdapter.cpp
sergey-chernikov: mostly for GUI - to allow some more actions with auth leaf (and to show it in the list of wallets)

BlocksettleNetworkingLib/Adapters/WalletsAdapter.cpp
sergey-chernikov: added

BlocksettleNetworkingLib/Adapters/BlockchainAdapter.cpp
sergey-chernikov: added settings and pointer checks

BlocksettleNetworkingLib/Adapters/WalletsAdapter.cpp
goatpig: Ideally it should be renamed to TxCount. The abbreviation txn is often used to mean transaction.

BlocksettleNetworkingLib/Adapters/WalletsAdapter.cpp
```diff
@@ -0,0 +1,2501 @@
...
+   return pushFill(envResp);
+}
+
+bool WalletsAdapter::processUnreserveUTXOs(const WalletsMessage_ReservationKey& request)
+{
+   utxoResMgr_->unreserve(request.id(), request.sub_id());
+   return true;
+}
+
+bool WalletsAdapter::processPayin(const bs::message::Envelope& env
```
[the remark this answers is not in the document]
goatpig: Let's plan for another PR to move auth, cc, futures and settlement logic into a set of dedicated "blocksettle services" folder/files. Anything related to login flow should probably go there too.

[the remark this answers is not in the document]
sergey-chernikov: Settlement logic already has its own adapter in the terminal. But generally I agree it should be finer granulated.

BlocksettleNetworkingLib/Adapters/WalletsAdapter.cpp
sergey-chernikov: renamed

Blocksettle_proto/Communication/common.proto
```diff
@@ -0,0 +1,952 @@
...
+   }
+
+   message WalletIDs
+   {
+      repeated string wallet_ids = 1;
+   }
+
+   message AddressTxCountResponse
+   {
+      message AddressTxNs {
```
goatpig, line 171: This should nested type should be renamed to count as well

goatpig (approved):
utACK

PR #2165, opened 2021-02-08
additional fields for future trades

sergey-chernikov:


sergey-chernikov:
Closing due to phantom conflict which doesn't exist. Use https://github.com/BlockSettle/common/pull/2169 instead

PR #2166, opened 2021-02-09
Bitfinex integration

Ation:


BlocksettleNetworkingLib/SslDataConnection.cpp
```diff
@@ -52,6 +52,14 @@ SslDataConnection::~SslDataConnection()
    closeConnection();
 }
 
+bool SslDataConnection::openConnectionWithPath(const std::string& host, const std::string& path
```
sergey-chernikov, line 55: Isn't it better to put `path` before `host` or after the `port`? host:port params usually go together

sergey-chernikov:
Do we need ca-bundle.crt.inc to be included to common? Didn't notice its usage in this PR.

BlocksettleNetworkingLib/SslDataConnection.cpp
Ation: agree. fixed

PR #2177, opened 2021-02-16
Price book updates

Ation:
- definitions required for PB request 
- replace system_clock in adapter bus with steady_clock to protect timers from system clock changes.

### DIFF
--- a/BlocksettleNetworkingLib/Message/Envelope.h
+++ b/BlocksettleNetworkingLib/Message/Envelope.h
@@ -19,7 +19,12 @@
 namespace bs {
    namespace message {
       using UserValue = int;
-      using TimeStamp = std::chrono::time_point<std::chrono::system_clock>;
+      using bus_clock = std::chrono::steady_clock;
+
+      static_assert(bus_clock::is_steady, "Should be steady clock");
+      static_assert(bus_clock::period::den >= 10000, "Should be at least ms");
+
+      using TimeStamp = std::chrono::time_point<bus_clock>;
 
       class User
       {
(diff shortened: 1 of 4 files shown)


BlocksettleNetworkingLib/Message/Envelope.h
```diff
@@ -19,7 +19,12 @@
 namespace bs {
    namespace message {
       using UserValue = int;
-      using TimeStamp = std::chrono::time_point<std::chrono::system_clock>;
+      using bus_clock = std::chrono::steady_clock;
```
sergey-chernikov, line 22: Nice change, but it should go to terminal_arch_refactor branch first (then merged to futures_dev). Please consider splitting this PR.

Blocksettle_proto/Communication/bs_server.proto
```diff
@@ -1578,6 +1583,7 @@ message PriceStreamManaging
       {
          string   owner_entity_name = 1;
          double   price             = 2;
+         int64    timestamp         = 3;
```
sergey-chernikov, line 1586: Why int64 and not uint64? Isn't timestamp always an absolute unix time?

BlocksettleNetworkingLib/Message/Envelope.h
Ation: You could cherry pick 2 commits. They are isolated from other changes. Same in PB PR, this change is an isolated adapter.

Blocksettle_proto/Communication/bs_server.proto
Ation: fixed

sergey-chernikov (approved):

PR #2202, opened 2021-03-09
Address API

Ation:


Blocksettle_proto/Communication/bs_api_server.proto
```diff
@@ -326,30 +344,58 @@ message Response
         string error_msg = 1;
     }
 
+    message SubmitAddress
+    {
+        string  address         = 1;
+        bool    success         = 2;
+        string  error_msg       = 3;
+        int64   added_timestamp = 4;
```
sergey-chernikov, line 352: timestamps are better to put as unsigned type, I think

Blocksettle_proto/Communication/bs_types.proto
```diff
@@ -117,3 +117,29 @@ enum ApiKeyError
     API_KEY_ERROR_UNKNOWN_IP_ADDR = 2;
     API_KEY_ERROR_SERVER_ERROR = 3;
 }
+
+message WhitelistedAddressRequest
```
sergey-chernikov, line 121: Why to put those to bs_types? I thought it was mostly destined for enums and really generic types. Even if putting here, IMO, better to make it more universal. I.e. not announce what you're gonna do with the address (whitelist or delist or update, or anything else) - so name it just AddressRequest/Response.

Ation: Agree that naming is not good. Goal I was trying to achieve was simple:
 - have 2 different message definitions, that could explicitly define that whitelisted address when submitted must not have any timestamps, and response for address submit should have a timestamp.
 - definition should be shared between bs_server_api and bs_server_api_pb requests, to simplify request passing from endpoint to backend.

sergey-chernikov: Ok, then I believe it can easily go to bs_api_server.proto out of bs_types.proto, as bs_api_server is included in bs_api_server_pb.

sergey-chernikov (approved):

PR #2228, opened 2021-03-29
Futures dev

sergey-chernikov:


BlocksettleNetworkingLib/HeadlessContainer.cpp
```diff
@@ -459,6 +457,14 @@ bs::signer::RequestId HeadlessContainer::signTXRequest(const bs::core::wallet::T
    case TXSignMode::Partial:
       packet.set_type(headless::SignPartialTXRequestType);
       break;
+
+   case TXSignMode::AutoSign:
```
goatpig, line 461: Making a note here: autosign will be deprecated as we're moving to an API model for dealing, which will put the burden of passphrase/credential management on the API client. There is no need for this logic in the terminal anymore and the related GUI element will be pruned out too.

goatpig (approved):
utACK

PR #2245, opened 2021-04-14
Add statements and charting to API server

pavel-kokolemin:


Blocksettle_proto/Communication/bs_api_server.proto
```diff
@@ -592,6 +605,19 @@ message Response
...
+    message GetStatementDates
+    {
+        bool loaded = 1;
+        repeated string dates = 2;
+    }
+
+    message GetStatements
+    {
+        bool loaded = 1;
+        string html = 2;
```
Ation, line 617: Very unclear field name. Can you provide a reference for that API call?

pavel-kokolemin: sure, here it's:
api_doc

PR #2253, opened 2021-04-27
Futures charting

Ation:
- remove EOD for futures
- add live cutoff updates to charting
- add new intervals and req/rep for futures OHLC


Blocksettle_proto/Communication/market_data_history.proto
```diff
@@ -16,79 +16,121 @@ import "trade_history.proto";
...
+   int64    lesser_then = 4;
+   uint64   request_id  = 5;
+};
+
+message OhlcFuturesRequest
+{
+   string                  product     = 1;
+   FuturesCandleIntervals  interval    = 2;
+   uint32                  count       = 3;
+   int64                   lesser_then = 4;
```
sergey-chernikov, line 87: typo? (then)
what is this field meaning in general, BTW?

Blocksettle_proto/Communication/market_data_history.proto
```diff
@@ -16,79 +16,121 @@ import "trade_history.proto";
...
-		Blocksettle.Communication.TradeHistory.TradeHistoryTradeType type = 2;
-	}
+   message Product{
+      string product = 1;
+      Blocksettle.Communication.TradeHistory.TradeHistoryTradeType type = 2;
+   }
     repeated Product products = 1;
-};
\ No newline at end of file
+};
```
sergey-chernikov, line 136: do we need colons (;) for proto syntax?

Blocksettle_proto/Communication/market_data_history.proto
Ation: Yes. Typo.
This is to load candles with timestamp lesser than lesser_than

Blocksettle_proto/Communication/market_data_history.proto
Ation: no, we don't need them. just a typing habbit.

Blocksettle_proto/Communication/market_data_history.proto
sergey-chernikov: ok, IMO better to specify also that it's a timestamp/unixtime then (or in comment)
In addition, if it's unixtime, it can only be unsigned type, I think.

sergey-chernikov (approved):

PR #2262, opened 2021-05-13
Message bus envelope rework

sergey-chernikov:


BlocksettleNetworkingLib/Message/Bus.cpp
```diff
@@ -90,7 +92,7 @@ std::vector<std::shared_ptr<bs::message::Adapter>> Router::process(const bs::mes
 {
    std::set<std::shared_ptr<bs::message::Adapter>> result;
    if (supervisor_ && !supervisor_->process(env)) {
-      logger_->debug("[Router::process] msg {} skipped by supervisor", env.id);
+      logger_->info("[Router::process] msg #{} skipped by supervisor", env.id());
```
goatpig, line 95: please renamed "skipped" to "intercepted", skipped is ambiguous (could mean supervisor ignored the message).

BlocksettleNetworkingLib/Message/Bus.cpp
```diff
@@ -137,14 +138,27 @@ void Router::reset()
...
 }
 
+bool bs::message::QueueInterface::isValid(const Envelope& env)
+{
+   if (!env.sender) {
+      return false;
+   }
+   const auto& itDeferred = deferredIds_.find(env.id());
+   if (itDeferred != deferredIds_.end()) {
+      deferredIds_.erase(itDeferred);
```
goatpig, line 156: why would a call to isValid() remove the id from deferredIds_? Shouldn't that only happen after a deffered message is processed?

BlocksettleNetworkingLib/Message/Envelope.h
```diff
@@ -73,15 +73,51 @@ namespace bs {
...
+         void setId(SeqId id)
+         {
+            id_ = id;
+            if (!foreignId_) {
+               foreignId_ = id;
+            }
+         }
+
+         SeqId foreignId() const { return foreignId_; }
+         void setForeignId(SeqId id) { foreignId_ = id; }
```
goatpig, line 109: What is this for? From what I can see, it's only ever used in conjunction with rabbitmq correlationID

BlocksettleNetworkingLib/Message/Envelope.h
```diff
@@ -73,15 +73,51 @@ namespace bs {
...
+
          std::shared_ptr<User>   sender;
          std::shared_ptr<User>   receiver;
          TimeStamp   posted;
          TimeStamp   executeAt;
          std::string message;
-         bool  request{ false };
+         SeqId responseId{ 0 };  // should be set in reply and for special flags
```
goatpig, line 116: Does this have to be public? I'd rather it is private and set at ctor only

BlocksettleNetworkingLib/Message/Envelope.h
sergey-chernikov: Not only RMQ-related. Any envelope gatewayed from another queue should have this set to non-local id. This is used already in Genoa adapter local bus.

BlocksettleNetworkingLib/Message/Bus.cpp
sergey-chernikov: No, deferred id is added again upon next message defer. It's done here just for the sake of performance - to avoid finding it again in the main processing loop.

BlocksettleNetworkingLib/Message/Bus.cpp
sergey-chernikov: "Skipped" here means that it's then not pushed back into the queue. Not sure what's the correct wording here. IMO, "intercepted" is not good, too, as it doesn't disclose what happens after "interception".

goatpig: supervisor skipped A: the supervisor did not supervise A
supervisor intercepted A: the supervisor stopped A on its course

I guess "consume" could work?

sergey-chernikov: IMO "seized" fits better

goatpig (approved):
utACK

PR #2267, opened 2021-05-20
Relay adapter implementation (stand-alone and sub-classed)

sergey-chernikov:


BlocksettleNetworkingLib/Message/Adapter.cpp
```diff
@@ -71,3 +71,67 @@ bool PipeAdapter::process(const Envelope &env)
...
+   }
+   queues_.insert(queue);
+   for (const auto& user : queue->supportedReceivers()) {
+      queueByUser_[user] = queue;
+   }
+}
+
+bool RelayAdapter::process(const Envelope& env)
+{
+   if (!fallbackUser_) {
```
goatpig, line 102: This sanity check is used in 3 seperate places, can you turn into in a "isInitialized" private method instead?

sergey-chernikov: done

goatpig (approved):
utACK

PR #2277, opened 2021-06-30
Market order updates

Ation:


Ation:
> Why don't use the "Response" flag? It's destined exactly for that.
Mostly because this is not a response. Response as a term always imply that there was a request. Update is more appropriate name.
>Also, why -4, not -3? The flags are not intended to combine.
If they are not intended to combine they should not be called flags. Until they are called flags that should look and act as flags.

sergey-chernikov:
Please read the response flag description carefully. It even has almost the same comment, as your new flag. Response is a non-request, and is used the same way as your new flag (semantically). I see no reason to add a new flag which does exactly the same.
Uncombinable flags - why not? You can set only one. Even in your case of -4, I'm pretty sure you can't combine it either with any other flag.

goatpig (approved):
tACK

PR #2285, opened 2021-07-13
Fix history

pavel-kokolemin:


Ation:
Please remove count form history request in definitions as well

PR #2295, opened 2021-07-30
Eliminate Envelope copy constructor

sergey-chernikov:


BlocksettleNetworkingLib/Message/Envelope.h
```diff
@@ -80,16 +80,17 @@ namespace bs {
...
-      struct Envelope
+      class Envelope
       {
          friend class QueueInterface;
          friend class Router;
          friend class RelayAdapter;
 
+      public:
          Envelope() = default;
          ~Envelope() noexcept = default;
```
Ation, line 95: One of the requests was to enforce rule-of-5 to avoid some warnings. I guess we could just remote explicit declarations of default ctors. Not sure though.

sergey-chernikov: Removed, though didn't make a test build

sergey-chernikov: Tested with BS_Private/ses_manager_roll, and default ctor was added back, as build fails without it.

goatpig (commented):
utACK

PR #2324, opened 2021-09-23
Processing thread fixes

sergey-chernikov:
Mainly for Dealer usage, not used widely in the other code.

### DIFF
--- a/CommonLib/ProcessingThread.h
+++ b/CommonLib/ProcessingThread.h
@@ -13,6 +13,7 @@
 #define __PROCESSING_THREAD_H__
 
 #include <atomic>
+#include <chrono>
 #include <memory>
 #include <queue>
 #include <thread>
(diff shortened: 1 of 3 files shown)


CommonLib/ProcessingThread.h
```diff
@@ -79,41 +89,62 @@ class ProcessingThread
...
+                     }
+                     packets.emplace_back(std::move(packet));
+                  }
+                  delayedPackets_.erase(itDelayed);
+               }
             }
 
-            if (pendingPackets_.empty()) {
-               pendingPacketsEvent_.ResetEvent();
+            while (!pendingPackets_.empty()) {
```
goatpig, line 124: Why the busy loop? This kind of approach suggests other threads may append pendingPackets_ while this thread is consuming its content, but a lock is acquired on pendingPacketLock_ at the begining of this scope.

sergey-chernikov: Not exactly. This loop just suggests batch processing if multiple packets were added before lock was acquired.

goatpig: Why does this require a while loop though? What does it provide that this doesnt:
packets.insert(packets.end(), pendingPackets_.begin(), pendingPackets.end())

sergey-chernikov: `pendingPackets_.clear();`
but anyway this is another way of doing the same thing

goatpig (approved):
tACK

PR #2325, opened 2021-09-27
Few fixes to ManualResetEvent

Ation:
- no need to use atomic, if we use mutex
- each notify_all return control from cv.wait(). Update implementation to avoid return from WaitFor until event actually set


goatpig (approved):
utACK

PR #2344, opened 2021-11-17
Add cjose build scripts

sergey-chernikov:


build_scripts/cjose_settings.py
```diff
@@ -0,0 +1,76 @@
+#
+#
+# ***********************************************************************************
+# * Copyright (C) 2020 - 2021, BlockSettle AB
```
Ation, line 4: it is 2021

build_scripts/cjose_settings.py
```diff
@@ -0,0 +1,76 @@
...
+        return os.path.join(self._project_settings.get_common_build_dir(), 'cjose')
+
+    def is_archive(self):
+        return True
+
+    def config(self):
+        command = ['cmake',
+            self.get_source_dir(),
+            '-DOPENSSL_INCLUDE_DIR=' + os.path.join(self._project_settings.get_common_build_dir(), 'OpenSSL/include'),
+            '-DJANSSON_INCLUDE_DIR=' + os.path.join(self._project_settings.get_common_build_dir(), 'jansson/include'),
```
Ation, line 48: Ideally it should be getting install dir directly from corresponding lib settings object. Qt and WS settings could be used as an example:
self.openssl = OpenSslSettings(settings)
os.path.join(self.openssl.get_install_dir(),'include'))

build_scripts/jansson_settings.py
```diff
@@ -0,0 +1,74 @@
...
+
+from component_configurator import Configurator
+
+class JanssonSettings(Configurator):
+    def __init__(self, settings):
+        Configurator.__init__(self, settings)
+        self._version = '2.14'
+        self._script_revision = '1'
+        self._package_name = 'jansson-master'
+        self._package_url = 'https://github.com/akheron/jansson/archive/refs/heads/master.zip'
```
Ation, line 24: please use specific revision

build_scripts/cjose_settings.py
```diff
@@ -0,0 +1,76 @@
...
+
+from component_configurator import Configurator
+
+class CJOSE(Configurator):
+    def __init__(self, settings):
+        Configurator.__init__(self, settings)
+        self._version = '0.6.1'
+        self._script_revision = '1'
+        self._package_name = 'cjose-master'
+        self._package_url = 'https://github.com/sergey-chernikov/cjose/archive/refs/heads/master.zip'
```
Ation, line 24: please use fixed revision

build_scripts/jansson_settings.py
```diff
@@ -0,0 +1,74 @@
+#
+#
+# ***********************************************************************************
+# * Copyright (C) 2020 - 2021, BlockSettle AB
```
Ation, line 4: 2020

build_scripts/cjose_settings.py
sergey-chernikov: it's the only revision supporting cmake builds, and as you can see it's my fork, no other revisions are expected in the nearest future (the original lib is not updated for 6 years), I don't intend to create any revisions of it

build_scripts/jansson_settings.py
sergey-chernikov: I will if you point me out to how to get this link with revision of that lib from github, I can see a link only to master archive, and no releases tab. Though getting the latest master doesn't seem a bad idea except that we need to increase script revision to update the build.

build_scripts/cjose_settings.py
sergey-chernikov: Changed, thank you

build_scripts/cjose_settings.py
Ation: `self._version = '3f77bc82e078c6aa15cbffa796fb2d700612ecf4'
self._package_url = 'https://github.com/sergey-chernikov/cjose/archive/' + self._version + '.zip'
`

build_scripts/jansson_settings.py
Ation: self._package_url = 'https://github.com/akheron/jansson/archive/refs/tags/' + self._version + '.zip'

Ation (approved):

PR #2350, opened 2021-12-07
add login client adapter

sergey-chernikov:


Ation:
New files might need copyrights, since this is a public repo

goatpig (approved):
tACK

PR #2360, opened 2022-02-01
Merge with the latest Armory/dev

sergey-chernikov:


WalletsLib/CoreHDGroup.cpp
```diff
@@ -312,6 +318,21 @@ void bs::core::hd::HWGroup::initLeafXpub(
    BIP32_Node newPubNode;
    newPubNode.initFromBase58(SecureBinaryData::fromString(xpub));
 
+   //no derivation path is passed to the account, it will use the pub root as is
+   auto accTypePtr = AccountType_BIP32::makeFromDerPaths(0, { });  //FIXME: seed fp
```
goatpig, line 322: This needs the node's fingerprint (instead of 0 as the first arg):

>   auto fingerprint = newPubNode.getThisFingerprint();
>   auto accTypePtr = AccountType_BIP32::makeFromDerPaths(fingerprint, { pathInt });

WalletsLib/CoreHDGroup.cpp
```diff
@@ -438,24 +448,26 @@ void hd::AuthGroup::initLeaf(std::shared_ptr<hd::Leaf> &leaf
    if (salt_.getSize() != 32) {
       throw AccountException("empty auth group salt");
    }
-   auto accTypePtr = std::make_shared<AccountType_BIP32_Salted>(pathInt, salt_);
+   auto accTypePtr = AccountType_BIP32_Salted::makeFromDerPaths(0, { pathInt }, salt_);   //FIXME: seed fp
```
goatpig, line 451: seed fingerprint can be had from the wallet:

>   auto rootBip32 = std::dynamic_pointer_cast<AssetEntry_BIP32Root>(walletPtr_->getRoot());
>   auto seedFingerprint = rootBip32->getSeedFingerprint(true);

WalletsLib/CoreHDGroup.cpp
```diff
@@ -438,24 +448,26 @@ void hd::AuthGroup::initLeaf(std::shared_ptr<hd::Leaf> &leaf
...
+      accTypePtr->setInnerAccountID(hd::Leaf::addrTypeInternal_);
    }
    else {
       //ext only address account uses the same asset account for both outer and
       //inner chains
       accTypePtr->setNodes({ hd::Leaf::addrTypeExternal_ });
-      accTypePtr->setOuterAccountID(WRITE_UINT32_BE(hd::Leaf::addrTypeExternal_));
-      accTypePtr->setInnerAccountID(WRITE_UINT32_BE(hd::Leaf::addrTypeExternal_));
+      accTypePtr->setOuterAccountID(hd::Leaf::addrTypeExternal_);
+      accTypePtr->setInnerAccountID(hd::Leaf::addrTypeExternal_);
```
goatpig, line 464: It isnt required to explicitly set the inner account anymore, it will default to the outer account when missing

WalletsLib/CoreHDGroup.cpp
```diff
@@ -470,7 +482,7 @@ void hd::AuthGroup::initLeaf(std::shared_ptr<hd::Leaf> &leaf
    auto accID = walletPtr_->createBIP32Account(accTypePtr);
 
    authLeafPtr->setPath(path);
-   authLeafPtr->init(walletPtr_, accID);
+   authLeafPtr->init(walletPtr_, accID.getAddressAccountKey());
```
goatpig, line 485: It would be preferable to carry AddressAccountId instead of AddressAccountKey (you will have to reconstruct the id every time to query the underlying AddressAccount from the corresponding AssetWallet otherwise)

WalletsLib/CoreHDLeaf.cpp
```diff
@@ -31,7 +35,7 @@ hd::Leaf::~Leaf()
 
 void hd::Leaf::init(
    std::shared_ptr<AssetWallet_Single> walletPtr,
-   const BinaryData& addrAccId)
+   const Armory::Wallets::AccountKeyType& addrAccId)
 {
    reset();
    auto accPtr = walletPtr->getAccountForID(addrAccId);
```
goatpig, line 41: There is no native getAccountForID(const Armory::Wallets::AccountKeyType&) method in AssetWallet, this is implicitly relying on the AddressAccountID(AddressAccountKey) ctor. Better use AddressAccountID explicitly.

WalletsLib/CoreHDLeaf.cpp
```diff
@@ -68,21 +72,20 @@ void hd::Leaf::reset()
...
 {
    return walletPtr_->beginSubDBTransaction(BS_WALLET_DBNAME, false);
 }
 
 std::string hd::Leaf::walletId() const
 {
    if (walletId_.empty()) {
-
-      walletId_ = wallet::computeID(getRootId()).toBinStr();
+      walletId_ = AddressAccountId(getRootId()).toHexStr();
```
goatpig, line 88: Same general remark about using AddressAccountId directly. Here the id is grabbed from the AddressAccount, converted to a key (getRootId()), then converted back to an id, to finally get a hex string.

WalletsLib/CoreHDLeaf.cpp
```diff
@@ -343,9 +325,7 @@ BinaryData hd::Leaf::serialize() const
    bw.put_uint32_t(LEAF_KEY);
 
    //address account id
-   auto&& rootID = getRootId();
-   bw.put_var_int(rootID.getSize());
-   bw.put_BinaryData(rootID);
+   bw.put_int32_t(getRootId());
```
goatpig, line 328: Changing the serialization format will likely break wallet files created with the previous version

WalletsLib/CoreHDGroup.cpp
```diff
@@ -312,36 +317,46 @@ void bs::core::hd::HWGroup::initLeafXpub(
...
+   }
    accTypePtr->setDefaultAddressType(leaf->defaultAddressType());
    accTypePtr->setAddressLookup(lookup);
-   accTypePtr->setOuterAccountID(WRITE_UINT32_BE(*nodes.begin()));
-   accTypePtr->setInnerAccountID(WRITE_UINT32_BE(*nodes.rbegin()));
+   accTypePtr->setOuterAccountID(*nodes.begin());
+   accTypePtr->setInnerAccountID(*nodes.rbegin());
    accTypePtr->setMain(true);
 
+#if 0 //FIXME
```
goatpig, line 340: do we have no test that uses this code (create WO account from xpub)? Shouldn't we try to restore the functionality?

sergey-chernikov: There're no corresponding test in BS_Private, but maybe in the terminal. I need help fixing this anyway.

goatpig (approved):
tACK

PR #2390, opened 2022-04-19
Decimal types ser/deser refactor

sergey-chernikov:


Ation:
noexcept qualifiers lost from ctors definitions.

sergey-chernikov:
Should we get rid of `noexcept`, or default-init in case of mismatch?

goatpig (approved):
tACK

PR #2452, opened 2022-07-21
[LEV-2343] Deprecate QNetworkAccessManager

sergey-chernikov:


BlocksettleNetworkingLib/SettingsParser.cpp
```diff
@@ -11,20 +11,18 @@
...
    : logger_(logger)
 {
-   addParam(SettingsFile, "settings_file", settingsFile.c_str(), "Path to settings file");
+   addParam(SettingsFile, "settings_file", settingsFile, "Path to settings file");
 }
 
-bool SettingsParser::LoadSettings(const QStringList& argList)
+bool SettingsParser::LoadSettings(int argc, const char** argv)
 {
+#if 0 //TODO: implement command-line+ini settings parser
```
Ation, line 25: Can we just remove that class?

sergey-chernikov: We probably can, but it's used in BsDbMon (also turned off from building temporarily).

Ation: We could remove them both :)

goatpig (approved):
utACK

PR #2464, opened 2022-08-22
[LEV-2351] Added BIP39 lib

sergey-chernikov:


WalletsLib/bip39/lib/picosha2.h
```diff
@@ -0,0 +1,377 @@
+/*
```
goatpig, line 1: do we really need a 5th different implementation of sha256?

goatpig (approved):
utACK

PR #2468, opened 2022-08-26
[LEV-2512] wallet rescan improvements

sergey-chernikov:


BlocksettleNetworkingLib/Adapters/WalletsAdapter.cpp
```diff
@@ -888,14 +978,66 @@ void WalletsAdapter::sendBalanceRequest(const std::string &walletId)
...
+   sendWalletChanged(hdWalletId);
+}
+
+void WalletsAdapter::scan(const std::shared_ptr<bs::sync::hd::Wallet>& hdWallet
+   , const std::shared_ptr<bs::sync::Wallet>& wallet
+   , const std::function<void(bs::sync::SyncState)>& cb)
+{
+   logger_->debug("[WalletsAdapter::scan] {}", wallet->walletId());
+   wallets_[wallet->walletId()] = wallet;
+   scanWallet(wallet, true);
```
goatpig, line 1039: why is this called twice?

sergey-chernikov: external and internal addresses (2nd arg)

goatpig: A scan operation is several minutes long at least. It also cannot happen in parallel. Queuing multiple set of addresses to be rescanned independently is not an efficient use of armorydb. You should batch the rescan of all concerned wallets in a single call.

goatpig (approved):
utACK

